PR #29, opened 2017-03-31
Non-zero exit status when fail

kitsuyui:
Issue #26


horike37:
@kitsuyui 
Thank you for the greate addition. just merged it:blush:

PR #32, opened 2017-04-11
typo

TrentBartlem:


horike37:
@TrentBartlem 
Nice catch! Thank you for sending PR!

PR #50, opened 2017-06-05
Custom naming for serverless-step-functions

icarus-sullivan:
Allow users to provide their own stepMachine naming convention.

_example_ 
```
functions:
  hello:
    handler: handler.hello

stepFunctions:
  stateMachines:
    hellostepfunc:
      Name: hellostepfunc-${opt:stage}
      definition:
        <your definition>

plugins:
  - serverless-step-functions
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,3 +154,26 @@ plugins:
   - serverless-step-functions
 ```
 
+### Adding a custom name for a stateMachine
+In case you need to interpolate a specific stage or service layer variable as the
+stateMachines name you can add a `Name` property to your yaml.
+
+```yml
+service: messager
+
+functions:
+  sendMessage:
+    handler: handler.sendMessage
+
+stepFunctions:
+  stateMachines:
+    sendMessageFunc:
+      Name: sendMessageFunc-${self:custom.service}-${opt:stage}
+      definition:
+        <your definition>
+
+plugins:
+  - serverless-step-functions
+```
+
+Please note, that during normalization some characters will be changed to adhere to CloudFormation templates.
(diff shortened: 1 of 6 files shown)


icarus-sullivan:
Sorry it took a while to get the lint/tests necessary updates.

justinmchase:
First off 👏 

A few questions though:

In your example code above you have this line:
```
statemachine_arn: ${self:resources.Outputs.HelloStepfunc.Value}
```

But you have omitted the Resources->Outputs section from the example. Was that just an oversight or does this PR automatically add step function arns to outputs as well?

Also, for your name you have `Name: hellostepfunc-${opt:stage}` but the actual name/arn will still have the random string attached correct? 

e.g.
```
hellostepfunc-prod-QWUYMJP6O0QH
```

justinmchase:
I would recommend adding this field to the main Readme.md also:

https://github.com/horike37/serverless-step-functions/blob/master/README.md#setup

Maybe also update the Tips example at the bottom.

icarus-sullivan:
** I'm on mobile at the moment so my answer may be short and have spelling errors. 

The leftover `step_function` environment variable was a copy and paste from an example on these threads. However it will still work even with a custom name. 

Yes, CloudFormation appends the step functions with random characters. I can't link the docs right now but there is a Get:Name function (or something similar) that looks like it ignores the trailing characters when querying for step functions. I haven't tested that yet though.

horike37:
Suuuuuuper Cool @icarus-sullivan!!!
Thank you for sending this PR!! This looks good! 
I didn't come up with this approach.

I will test it locally. However I am busy in this week, please give me a little time.

horike37:
Thank you for this @icarus-sullivan !
I have tested locally and work it correctly! LGTM.
merged!

PR #56, opened 2017-06-27
Not indenting path and method results in error

biznickman:
If you copy the formatting of the README it results in an error `Cannot read property 'path' of null`. This fixed it for me.


horike37:
Nice catch @biznickman !
merged..

PR #109, opened 2018-03-07
Do not display endpoints if there are none

j0k3r:
Display nothing instead of:

```
Serverless StepFunctions OutPuts
endpoints:

```

Because that information is useless if user never configured endpoints.


j0k3r:
👍

PR #111, opened 2018-03-14
Add support for CloudFormation intrinsic function for role

alex-berger:
This PR adds support for CloudFormation intrinsic functions to the ```role``` property, such for example the following sytanx is supported:

```yaml
role: {"Fn::GetAtt": ["MyRoleId", "Arn"]}
role: {Fn::GetAtt: [MyRoleId, Arn]}
role:
   Fn::GetAtt:
   - MyRoleId
   - Arn
```
Having this feature in place, it is now possible to use (refer to) IAM roles which are defined in the ```Resources``` section of the ```serverless.yml``` file.

### DIFF
--- a/lib/deploy/stepFunctions/compileStateMachines.js
+++ b/lib/deploy/stepFunctions/compileStateMachines.js
@@ -2,6 +2,11 @@
 const _ = require('lodash');
 const BbPromise = require('bluebird');
 
+function isIntrinsic(obj) {
+  const isObject = typeof obj === 'object';
+  return isObject && Object.keys(obj).some((k) => k.startsWith('Fn::'));
+}
+
 module.exports = {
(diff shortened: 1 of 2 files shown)


lib/deploy/stepFunctions/compileStateMachines.js
```diff
@@ -2,6 +2,11 @@
 const _ = require('lodash');
 const BbPromise = require('bluebird');
 
+function isIntrinsic(obj) {
```
horike37: Why is this function outside `module.exports`? Would be good to be inside that.

JonChristensen:
Just a quick note that this feature is mentioned in the docs, but the syntax in the docs does not match the syntax shown by @alex-berger and therefore led me astray until I found this.

Issue #92, opened 2017-12-14
Support custom http request templates


dbaghdanov:
# Feature Proposal

## Description
I'd like be be able to customize the events / http / request / template.  I'd like to add my own content-type definitions, and even override the build in definitions for `application/json`.

For feature proposals:
* Currently
  * Currently a default Body Mapping Template is used which defines a mapping for `application/json` and `application/x-www-form-urlencoded`.
  * Currently the events / http / request / template is completely  ignored if it's included.
* Proposed
  * Let the script provide a request / template only if none is provided, or a match isn't found.
```
stepFunctions:
  stateMachines:
    myStateMachine:
      events:
        - http:
            path: orders/{id}
            method: POST
            request:
              template:
                application/xml: '{"input": "$input.body", "stateMachineArn": "arn:aws:states:region:account:stateMachine:myStateMachine"}'
              parameters:
                paths:
                  id: true
```

## Additional Data
* ***Serverless Framework Core Version you're using***: 
  * 1.23.0
* ***The Plugin Version you're using***:
  * 1.3.0
* ***Operating System***:
  * Windows 7

horike37:
Thanks @dbaghdanov for the feature request :+1: 
We just accepted the request.

hubertpompecki:
Is this being worked on at the moment? I would find it very useful and I am happy to have a go at implementing it (probably with some help) unless somebody else is looking at it already.

svdgraaf:
@horike37 is it possible to release this? :)

hubertpompecki:
Hi @horike37,

Are you aware of anyone working on this feature at the moment? I am looking at picking it up as I would find it very useful for one of the projects I am working on but I wouldn't want to duplicate somebody else's work.

Also, just to help with planning, how laborious do you think this is likely to be?

Finally, thanks for the great work on this plugin, we use it for some production services over here and it is awesome! :+1:

horike37:
Hi @hubertpompecki 
Thank you for jumping in this issue :smile: 

Anybody has not started this implementation yet so it would be great if you can take care of it.

>Also, just to help with planning, how laborious do you think this is likely to be?

Umm :thinking: 
We have stolen most part of this http event feature from sls core since actually these are mostly same feature, so you might be able to implement this with picking up some pieces around here.
https://github.com/serverless/serverless/blob/master/lib/plugins/aws/package/compile/events/apiGateway/lib/method/integration.js#L169

[...]

PR #122, opened 2018-05-25
Add support for custom body mapping templates

hubertpompecki:
Implements support for customizing body mapping templates as discussed in https://github.com/horike37/serverless-step-functions/issues/92.


horike37:
@hubertpompecki 
Thank you for implementing this feature!
LGTM 💯

PR #124, opened 2018-05-25
Support custom logicalId separate from name

dougmoscrop:
I think it would be better to just use the key as the logical id, but this is more backwards-compatible.

This allows specifying an id that is different than the name. This makes it more deterministic for purposes of Ref, etc.

```yml
stepFunctions:
  stateMachines:
    foo:
      id: MyStateMachine # this only needs to be unique within the template
      name: myStateMachine-${opt:stage} # this has to be unique across stages, etc.
```

In fact, where `key` means the stateMachines key for a given state machine, I think making the name default to "{service}-{key}-{stage}" and the logicalId to `{Key}StepFunctionStateMachine` is a very sensible default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,30 @@ plugins:
   - serverless-step-functions
 ```
 
+### Adding a custom logical id for a stateMachine
+You can use a custom logical id that is only unique within the stack as opposed to the name that needs to be unique globally. This can make referencing the state machine easier/simpler because you don't have to duplicate the interpolation logic everywhere you reference the state machine.
+
+```yml
+service: messager
+
+functions:
+  sendMessage:
+    handler: handler.sendMessage
+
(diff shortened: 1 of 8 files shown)


horike37:
Hi, @dougmoscrop thank you for implementing such as the awesome enhancement 🎉 
Just tested this PR and looks great to see working as expected 👍 
I would love to add this to the next release!

Just one more thing, could you document the explanation of id statement to README?

dougmoscrop:
Will do!

dougmoscrop:
I updated the README and rebased on the latest changes.

dougmoscrop:
@horike37 sorry for the delay, fixed the lint issues too. should be good to go.

horike37:
@dougmoscrop 
Thanks!

PR #131, opened 2018-07-11
Closes #130 Shared API Gateway and maxAge property support for HTTP Endpoints

vlewin:
Closes #130 Shared API Gateway support for HTTP Endpoints

PR #133, opened 2018-07-14
Enable API Gateway authorizers

vlewin:
https://serverless.com/framework/docs/providers/aws/events/apigateway/#http-endpoints-with-aws-iam-authorizers


vlewin:
This PR is based on #131

vlewin:
@horike37 Hello Takahiro, is there a reason why not to use a serverless internal libraries instead of copying files from serverless source?

horike37:
Hey @vlewin !
Thank you for implementing this feature! Really looking forward to seeing this will be live as well as #131!!
Let me know if you will be ready for review 😄 

> is there a reason why not to use a serverless internal libraries instead of copying files from serverless source?

Good point 👍 
sls internal libraries around events have not been considered as libraries which referenced from external plugins.
Of course, I understanding the best way that using those in terms of keeping codebase clean. However, If I use it, this plugin might be broken when releasing the serverless framework new version.

That being said, I have decided to get copying way for keeping it safe.

vlewin:
@horike37 Thank you for quick response. Maybe we can pin tested and supported version of serverless in package.json and update it once a new version is released. So we can test your plugin and if all green release it. As you can see from my PR serverless is under active development and this results in a lot of changes per minor version. I already tested your plugin with the libraries from lastest serverless version and did't detected any issues.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
+# 1.5.1(10.07.2018)
```
horike37: Could you delete it once?
This should be added while releasing a new version.

lib/deploy/events/apiGateway/cors.js
```diff
@@ -24,6 +25,16 @@ module.exports = {
         'Access-Control-Allow-Credentials': `'${config.allowCredentials}'`,
       };
 
+      // Enable CORS Max Age usage if set
+      if (_.has(config, 'maxAge')) {
```
horike37: maxAge feature is great!
However, does this need for authorizer?
If not, would be great if you can separate maxAge PR from this one.

lib/deploy/events/apiGateway/resources.js
```diff
@@ -6,65 +6,247 @@ const _ = require('lodash');
 module.exports = {
```
horike37: Are those changes of this file for share API feature?
If so, could you remove those from this PR?

lib/deploy/events/apiGateway/resources.test.js
```diff
@@ -70,15 +70,15 @@ describe('#compileResources()', () => {
         },
```
horike37: Same here. maybe those are for share API.

lib/deploy/events/apiGateway/restApi.js
```diff
@@ -5,17 +5,55 @@ const BbPromise = require('bluebird');
...
+      this.serverless.service.provider.apiGateway.restApiId) {
+      return BbPromise.resolve();
+    }
+
     this.apiGatewayRestApiLogicalId = this.provider.naming.getRestApiLogicalId();
 
+    let endpointType = 'EDGE';
+
+    if (this.serverless.service.provider.endpointType) {
+      const validEndpointTypes = ['REGIONAL', 'EDGE', 'PRIVATE'];
```
horike37: Selecting endpointType is great!
However, those are not for authorizer feature.
Could you separate PR for each feature?

lib/deploy/events/apiGateway/restApi.js
```diff
@@ -5,17 +5,55 @@ const BbPromise = require('bluebird');
...
         Properties: {
           Name: this.provider.naming.getApiGatewayName(),
+          EndpointConfiguration: {
+            Types: [endpointType],
+          },
         },
       },
     });
 
+    if (!_.isEmpty(this.serverless.service.provider.resourcePolicy)) {
```
horike37: Same here. this is resourcePolicy feature.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "serverless-step-functions",
-  "version": "1.5.0",
+  "version": "1.6.0",
```
horike37: Could you turn back to 1.5.0?
This should be updated while releasing a new version.

lib/deploy/events/apiGateway/resources.js
vlewin: Hi @horike37 thank you for review! Yes those changes are needed for mapping serveless API Gateway configuration to CloudFormation resource properties e.g.:

```yml
  apiGateway: # Optional API Gateway global config
    restApiId: n42gvt3whk # REST API resource ID. Default is generated by the framework
    restApiRootResourceId: 9f8qmgn5d6 # Root resource ID, represent as / path
    restApiResources: # List of existing resources that were created in the REST API. This is required or the stack will be conflicted
      '/playground': skvi34
```

vlewin:
Hi @horike37 thank you for review! You are right this PR includes changes from #131. Please review #131 Shared API Gateway pull request first. 

Once this #131 is merged to master we will see only the changes for authorizers. If you want to add authorizers support first I can open a separate pull request but our team need both changes in the next version. Otherwise we have to use my fork what I want to avoid.

I'm sorry for the confusion.

horike37:
Hi @vlewin 
Could you rebase this branch from the top of master so that we can see only the changes of authorizer feature? #131 was just merged

vlewin:
@horike37 JFYI: branch is rebased, now we should see only authorizer related changes

vlewin:
Should we start using serverless internal libraries?
https://github.com/horike37/serverless-step-functions/pull/133/files#diff-6d186b954a58d5bb740f73d84fe39073R12

If yes we have to rename/reassign a property:
https://github.com/horike37/serverless-step-functions/pull/133/files#diff-6d186b954a58d5bb740f73d84fe39073R110

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "serverless-step-functions",
-  "version": "1.5.1",
+  "version": "1.5.0",
```
horike37: Please revert it

horike37:
Hi @vlewin, thank you for fixing!
I tested locally but does not work with the following error.

```
$ sls deploy -v
Serverless: Packaging service...
Serverless: Excluding development dependencies...
 
  Type Error ---------------------------------------------
 
  Cannot read property 'replace' of undefined
 
     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.
 
  Get Support --------------------------------------------
     Docs:          docs.serverless.com
     Bugs:          github.com/serverless/serverless/issues
     Issues:        forum.serverless.com
 
  Your Environment Information -----------------------------
     OS:                     darwin
     Node Version:           8.4.0
     Serverless Version:     1.29.1
```

Here is my serverless.yml file.
```yaml
service: step-func

provider:
  name: aws
  runtime: nodejs6.10

functions:
  hello:
    handler: handler.hello
  authorizerFunc:
    handler: authorizer.js

stepFunctions:
  stateMachines:
    step-func:
      events:
        - http:
            path: gofunction
            method: GET
            authorizer: authorizerFunc
      name: my-${opt:stage}-statemachine
      definition:
        StartAt: Ingest
        States:
          Ingest:
            Type: Task
            Resource: arn:aws:lambda:#{AWS::Region}:#{AWS::AccountId}:function:${self:service}-${opt:stage}-hello
            End: true


plugins:
  - serverless-pseudo-parameters
  - serverless-step-functions

```

Any ideas why?

horike37:
@vlewin 
Thank you for the update!
LGTM 👍 💯 I will release a new version soon!
Awesome work!!!!

vlewin:
@horike37 Awesome! Thank you very much for review!

Issue #140, opened 2018-08-05
Bug in ApiGateway Template: apostrophe results in parsing error


marcelalburg:
# This is a Bug Report

It looks like in the stepfunctions api gateway template is the same error as here:  https://github.com/serverless/serverless/pull/2993/files

## Description
If an ' (apostrophe) inside the request, it will always fail. 

For bug reports:
* What went wrong?
JSON parsing wrong.
* What did you expect should have happened?
A valid JSON must be accepted and redirected to the StepFunction
* What was the config you used?
```
    ApiGatewayCall:
      name: ${self:service}-${opt:stage, self:provider.stage}-ApiGatewayCall
      timeout: 100 # optional, in seconds, default is 6

      events:
        - http:
            path: import
            method: POST
```
* What stacktrace or error message from your provider did you see?
`{"__type":"com.amazon.coral.service#SerializationException"}`

* ***Provider Error messages***:
```
curl -D - -XPOST -H "Content-Type: application/json; charset=utf-8" https://<endpoint>/dev/import -d "{\"name\":\"Palatto d'Inv\"}"
HTTP/2 400
content-type: application/json
content-length: 60
date: Sun, 05 Aug 2018 19:59:20 GMT
x-amzn-requestid: 0afcd14b-98ea-11e8-ad05-8f443798b4c1
x-amz-apigw-id: LKqL4HhyliAFoYQ=
x-amzn-trace-id: Root=1-5b675718-08e0041dcc5f96d971896ced
x-cache: Error from cloudfront
via: 1.1 affe26bf02a36a4a45ea1eb3ce2b4a62.cloudfront.net (CloudFront)
x-amz-cf-id: 7wbWVgMmsFwHFIFXhJ8S204WSsP2VGDUaRMZngL1nylrr50TqXZP7Q==

{"__type":"com.amazon.coral.service#SerializationException"}
```

PR #141, opened 2018-08-05
Fix ApiGateway parsing error with single Quotes

marcelalburg:
Hello i wrote this ticket #140 and try to fix the template.

I ~~**could NOT test**~~ this code changes, but i did it manually inside my api gateway template and it works.

the final template should look likes this

```
#set( $body = $util.escapeJavaScript($input.json('$')).replaceAll("\\'", "'") ) 

{"input": "$body","name": "$context.requestId","stateMachineArn":"arn:aws:states:eu-central-1:xxx:stateMachine-xxx}
```

I tested it with a deployment and it works as expected

### DIFF
--- a/lib/deploy/events/apiGateway/methods.js
+++ b/lib/deploy/events/apiGateway/methods.js
@@ -138,7 +138,7 @@ module.exports = {
       'Fn::Join': [
         '', [
           // eslint-disable-next-line max-len
-          "#set( $body = $util.escapeJavaScript($input.json('$')).replaceAll(\"\\'\", \"'\") ) \n\n",
+          "#set( $body = $util.escapeJavaScript($input.json('$')).replaceAll(\"\\\\'\", \"'\") ) \n\n",
           '{"input": "$body","name": "$context.requestId","stateMachineArn":"',
           {
             Ref: `${stateMachineLogicalId}`,
--- a/lib/deploy/events/apiGateway/methods.js
+++ b/lib/deploy/events/apiGateway/methods.js
@@ -137,7 +137,7 @@ module.exports = {
     return {
       'Fn::Join': [
         '', [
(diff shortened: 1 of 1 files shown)


marcelalburg:
Thanks for your message. I'm not sure how to write a test for this. Even there was no test for this line before. You just trust, that this regular expression works right ?

lendesk-pawel-cisek:
@horike37 I've encountered similar problem it is described here
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html $util.escapeJavaScript() description. 
This code is untestable as this just generates the string which is later pass to AWS.
So the test would be just checking if the string matches hardcoded value which is put here
https://github.com/horike37/serverless-step-functions/pull/141/files#diff-e4e581bf0aabc7eb38729ca0f4518128R141
```
    it('should return the default template for application/json when one is not given', () => {
      const httpWithoutRequestTemplate = {
        path: 'foo/bar1',
        method: 'post',
        request: {
          template: {
            'application/x-www-form-urlencoded': 'custom template',
          },
        },
      };
      const requestTemplates = serverlessStepFunctions
        .getMethodIntegration('stateMachine', undefined, httpWithoutRequestTemplate)
        .Properties.Integration.RequestTemplates;
      expect(requestTemplates['application/json']['Fn::Join'][1][2].Ref)
        .to.be.equal('StateMachineStepFunctionsStateMachine');
      expect(requestTemplates['application/json']['Fn::Join'][1][0])
        .to.be.equal("#set( $body = $util.escapeJavaScript($input.json('$')).replaceAll(\"\\\\'\",\"'\") ) \n\n");
    });
```
cc: @marcelalburg

theburningmonk:
Closing because this PR has been superseded by #235

PR #143, opened 2018-08-15
Do not use StateMachineName for cloudformation resource key names

### LINKED ISSUE #142: Do not use step function custom `name` as cloudformation resource key name
# This is a (Feature Proposal)

## Description

When using a custom `name` in the definition of a step function in serverless, this is obviously intended to populate the `StateMachineName` [property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html#cfn-stepfunctions-statemachine-definitionname). It does that sucessfully

It has the unfortunate side-effect of also changing the resource key name used in the cloudformation template. This is not desirable as it makes it impossible to reference the newly created step function in a `resources` block because the name will be transformed. I can not think of any advantage in changing the resource key name to this value.  

For example, take this partial example:

```yaml
stepFunctions:
  stateMachines:
    MyStepFunction:
      name: ${self:provider.environment.STACK_PREFIX}-${self:provider.environment.STAGE}-MyStepFunction
      definition:
        ...
```

Where `STACK_PREFIX="test-1"` and `STAGE="dev"`. The resulting name is `test-1-dev-MyStepFunction` which is exactly what I want `StateMachineName` to be set to. However this causes the cloudformation template to look like this:

```json
"TestDash1DashdevDashMyStepFunction": {
  "Type": "AWS::StepFunctions::StateMachine",
  "Properties":
    ...
```

Note the capitalisation of the first letter and the replacement of the dashes with "Dash". If I then wish to reference this step function in my `resources` block I can usually use the known pattern that is used to create the cloudformation key names, so I can reference them, but in this case I can't, because it is a transformation of the name property I supplied.

```yaml
resources:
  Resources:

    MyStepFunctionARN:
      Type: AWS::SSM::Parameter
      Properties:
        Name: /${self:provider.environment.STACK_PREFIX}/${self:provider.environment.STAGE}/MyStepFunctionARN
        Description: ARN of my step function
        Type: String
        Value:
          Ref: <??????????>
```

A `Ref` that would work is `TestDash1DashdevDashMyStepFunction`, but as far as I know there is no way to (within the serverless template) access the mechanism to transform `test-1-dev-MyStepFunction` into this.

As there is no advantage to renaming the key name (as it's entirely internal to the template and it already has a well defined way to define its own key name), I suggest simply leaving it as it would have been if `name` had never been specified: `MyStepFunctionStepFunctionsStateMachine`.

## Additional Data
* ***Serverless Framework Core Version you're using***: 1.29.2
* ***The Plugin Version you're using***: serverless-step-functions@1.6.1

dpwrussell:
This changes the cloudformation resource key names that are chosen when a custom `name` for the step function is defined.

Without this change, it is not possible to have a custom name for a step function while still being able to internally reference resources by a predictable key name. I think it is also consistent with the way that serverless works in general. For example, if a lambda function is given a name then it does not affect the cloudformation key name used.

```YAML
functions:
  myLambdaFunction:
    name: ${self:provider.environment.STACK_PREFIX}-${self:provider.environment.STAGE}-myLambdaFunction
    handler: handler.myLambda
```

Will yield a cloudformation key name of `MyLambdaFunctionLambdaFunction`.

It would seem to me that it is extremely important to remain consistent (and non-parameterised) within the serverless config and cloudformation template generation with respect to key names because of the extremely frequency with which it is necessary to use the `resources` block escape hatch.

#### Unit Tests
I have fixed the unit tests as they current exist, but if this is merged, then I think it is likely that several should be removed as they basically test nothing more than the non-custom name case now.

Resolves #142

### DIFF
--- a/lib/deploy/events/apiGateway/methods.test.js
+++ b/lib/deploy/events/apiGateway/methods.test.js
@@ -103,11 +103,11 @@ describe('#methods()', () => {
         .to.be.equal('StateMachineStepFunctionsStateMachine');
     });
 
-    it('should set custom stateMachinelogical ID to RequestTemplates when customName is set',
+    it('should set stateMachinelogical ID to RequestTemplates when customName is set',
     () => {
       expect(serverlessStepFunctions.getMethodIntegration('stateMachine', { name: 'custom' })
         .Properties.Integration.RequestTemplates['application/json']['Fn::Join'][1][2].Ref)
-        .to.be.equal('Custom');
+        .to.be.equal('StateMachineStepFunctionsStateMachine');
     });
 
     it('should set Access-Control-Allow-Origin header when cors is true',
@@ -138,12 +138,12 @@ describe('#methods()', () => {
         .to.be.equal('StateMachineStepFunctionsStateMachine');
     });
 
-    it('should set custom stateMachinelogical ID in default templates when customName is set',
+    it('should set stateMachinelogical ID in default templates when customName is set',
     () => {
       const requestTemplates = serverlessStepFunctions
         .getIntegrationRequestTemplates('stateMachine', { name: 'custom' });
       expect(requestTemplates['application/json']['Fn::Join'][1][2].Ref)
-        .to.be.equal('Custom');
+        .to.be.equal('StateMachineStepFunctionsStateMachine');
     });
 
     it('should return the default template for application/json when one is not given', () => {
--- a/lib/deploy/stepFunctions/compileStateMachines.test.js
+++ b/lib/deploy/stepFunctions/compileStateMachines.test.js
@@ -39,44 +39,44 @@ describe('#compileStateMachines', () => {
 
     expect(serverlessStepFunctions.serverless.service
       .provider.compiledCloudFormationTemplate.Resources
-      .StateMachineBeta1.Type
+      .MyStateMachine1StepFunctionsStateMachine.Type
     ).to.equal('AWS::StepFunctions::StateMachine');
     expect(serverlessStepFunctions.serverless.service
       .provider.compiledCloudFormationTemplate.Resources
-      .StateMachineBeta2.Type
+      .MyStateMachine2StepFunctionsStateMachine.Type
     ).to.equal('AWS::StepFunctions::StateMachine');
     expect(serverlessStepFunctions.serverless.service
       .provider.compiledCloudFormationTemplate.Resources
-      .StateMachineBeta1.Properties.DefinitionString
+      .MyStateMachine1StepFunctionsStateMachine.Properties.DefinitionString
     ).to.equal('"definition1"');
     expect(serverlessStepFunctions.serverless.service
       .provider.compiledCloudFormationTemplate.Resources
-      .StateMachineBeta2.Properties.DefinitionString
+      .MyStateMachine2StepFunctionsStateMachine.Properties.DefinitionString
     ).to.equal('"definition2"');
     expect(serverlessStepFunctions.serverless.service
       .provider.compiledCloudFormationTemplate.Resources
-      .StateMachineBeta1.Properties.RoleArn['Fn::GetAtt'][0]
+      .MyStateMachine1StepFunctionsStateMachine.Properties.RoleArn['Fn::GetAtt'][0]
     ).to.equal('IamRoleStateMachineExecution');
     expect(serverlessStepFunctions.serverless.service
       .provider.compiledCloudFormationTemplate.Resources
-      .StateMachineBeta2.Properties.RoleArn['Fn::GetAtt'][0]
+      .MyStateMachine2StepFunctionsStateMachine.Properties.RoleArn['Fn::GetAtt'][0]
     ).to.equal('IamRoleStateMachineExecution');
     expect(serverlessStepFunctions.serverless.service
       .provider.compiledCloudFormationTemplate.Resources
-      .StateMachineBeta1.DependsOn
+      .MyStateMachine1StepFunctionsStateMachine.DependsOn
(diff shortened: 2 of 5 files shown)


dpwrussell:
This is failing on Travis because the functions which control the key names now have an unused parameter:

```
  getStateMachineLogicalId(stateMachineName, stateMachine) {
    return `${this.provider.naming
      .getNormalizedFunctionName(stateMachineName)}StepFunctionsStateMachine`;
  },

  getStateMachineOutputLogicalId(stateMachineName, stateMachine) {
    return `${this.provider.naming
      .getNormalizedFunctionName(stateMachineName)}StepFunctionsStateMachineArn`;
  },
```

If this gets the nod I am happy to go back and refactor that so that the functions have the second argument removed. I would have done this already, but this has a similar knock-on effect on other functions which call this one.

dpwrussell:
Hi,

This really is essential to make custom, but predictable resources. I am happy to rebase this off the latest if you are happy to merge this?

Thanks

xrendan:
Hey @horike37, 

I agree with @dpwrussell on this, this is essential to make a sensible serverless.yml file. Any status?

Thanks,
Brendan :)

horike37:
Thank you for working on this @dpwrussell and sorry for so long delay response.
I agree with what you stated. This was needed when `StateMachineName` property of `AWS::StepFunctions::StateMachine` has not been introduced but no need as of now.

I have reviewed your code and almost looks good. However, you removed [this feature](https://github.com/horike37/serverless-step-functions#user-content-adding-a-custom-logical-id-for-a-statemachine) as well. Please leave it as it is, otherwise a breaking change will be intoriduced.

dpwrussell:
@horike37 No worries, glad you're keen on fixing it. I'll try and get a chance to look at it this weekend.

dpwrussell:
@horike37 Sorry, it was Christmas, have just got to this now. I've done as you asked, which actually made the PR simpler as the refactoring was unnecessary.

I'd appreciate it if you would look over the tests in particular as I've adapted them to not handle name, but you might want to reword some of them. Let me know if you prefer to change the text of any of them if so and I will amend this PR before merging.

Cheers

horike37:
Thank you for doing that @dpwrussell 👍
I just tested and fresh deployment worked fine but it failed with following error message if there is existing deployment. This would be a breaking change, which means this PR would break deployed state machines. 

error message
``` shell
$ curl https://zzzzzzzzzz.execute-api.us-east-1.amazonaws.com/dev/hello
{"__type":"com.amazonaws.swf.service.v2.model#StateMachineDoesNotExist","message":"State Machine Does Not Exist: 'arn:aws:states:us-east-1:xxxxxxxxxxxx:stateMachine:sususu'"}
```

my serverless.yml
``` yaml
service: step-func

provider:
  name: aws
  runtime: go1.x
  timeout: 30
  region: us-east-1

functions:
  hello:
    handler: create-task-fn

stepFunctions:
  stateMachines:
    hello:
      events:
        - http:
            path: hello
            method: GET
      name: sususu
      definition:
        Comment: "A Hello World example of the Amazon States Language using an AWS Lambda Function"
        StartAt: HelloWorld1
        States:
          HelloWorld1:
            Type: Task
            Resource: arn:aws:lambda:#{AWS::Region}:#{AWS::AccountId}:function:${self:service}-${opt:stage}-hello
            End: true

plugins:
  - serverless-pseudo-parameters
  - serverless-step-functions

```

Here is the step to reproduce
- deploy statemachines with the current version of serverless-step-functions
- update serverless-step-functions to this PR version
- re-deploy with the same serverless.yml

Any ideas why?

dpwrussell:
@horike37 Sorry for not getting back to you sooner. I've been very busy.

So, basically what happens is that when `serverless-step-functions` creates the template it hardcodes the ARN of the step function into the mapping template for the content types:

```json
#set( $body = $util.escapeJavaScript($input.json('$')) ) 

{"input": "$body","name": "$context.requestId","stateMachineArn":"arn:aws:states:us-east-1:292075781285:stateMachine:sususu"}
```

And when the update occurs, that does not get updated. This is not a problem unique to this PR, but seems to be a problem in the general case. If you deploy your above example then switch it to use a custom logical ID

```yaml
...
stepFunctions:
  stateMachines:
    hello:
      events:
        - http:
            path: hello
            method: GET
      name: sususu
      id: sususureid
      definition:
        Comment: "A Hello World example of the Amazon States Language using an AWS Lambda Function"
        StartAt: HelloWorld1
        States:
          HelloWorld1:
            Type: Task
            Resource: arn:aws:lambda:#{AWS::Region}:#{AWS::AccountId}:function:${self:service}-${opt:stage}-hello
            End: true
...
```

Then the hardcoded value in the mapping templates is not updated either.

Obviously that is more of a conscious change than simply updating the library, but I think probably that hardcoded value should be updated when the logical ID of the step function is updated in the general case, which would also fix the problem you are seeing. Is that possible?

horike37:
@dpwrussell 
Umm. 100% not agreed with that. Many users probably update this plugin without reading README or PR or release note. It might make many users confusion if doing that. I want to avoid it.

Anyway, I have done more deep debugging with the same situation but found weird behavior.
The statemachines have been deleted when the second deploy after updating this implementation.
Umm 🤔 The root cause might be AWS side.

cc: @theburningmonk

dpwrussell:
That is what I was saying, that this is more severe than a configuration change as it can happen without the user being aware of it.

My point is that changing the resource name (be it from this PR, or an explicit addition/change of `id`) already has a problem, which if addressed elsewhere, would mean this PR could be applied without being a breaking change.

siovaneDAZN:
Has this issue been solved? Any chance to use the plugin  in the same way as lambdas?

Issue #146, opened 2018-08-28
IamRoleStateMachineExecution - Policy statement must contain resources


spongeboy:
# This is a Bug Report

## Description

Also reported in [Serverless forums](https://forum.serverless.com/t/iamrolestatemachineexecution-policy-statement-must-contain-resources/5594)

Using version 1.7.0 of plugin caused a deployment error of
`IamRoleStateMachineExecution - Policy statement must contain resources`

We resolved the issue by downgrading back to version 1.6.1.

Apologies, but at the moment I am unable to supply a sample config, as I need strip out sensitive information.

Our step function uses parallel steps, and the same function is referenced in multiple steps.

## Environment
OS:                     linux
Node Version:           8.9.4
Serverless Version:     1.25.0

ACSchott526:
There is a bug in the regex here: https://github.com/horike37/serverless-step-functions/commit/f55a54758f7638ebd06391c8b83e9b333d8727e2#diff-60425c3006015f918176a0be211f414bR15

The regex does not allow `*` or any numbers which are probably apart of most people's function arn's (wildcard and account id/region, respectively). As a result, the resources field is set to an empty array and CloudFormation complains.

kabo:
We're still having this issue in `1.7.2`.

Our `Resource` line in `serverless.yml`:

```yaml
Resource: "arn:aws:lambda:#{AWS::Region}:#{AWS::AccountId}:function:${self:service}-${self:custom.stage}-flow161cfetchFromSystemPlus"
```

uris77:
Has anyone solved this issue. I just bumped into it with version 1.8

bouiboui:
Many people are still affected by this:
https://forum.serverless.com/t/iamrolestatemachineexecution-policy-statement-must-contain-resources/5594/9

jimmyn:
I'm having the same issue. Had to revert to 1.6.1. Any update on this?

nakamine-tsc:
I have the same issue with version >= 1.7.2. But in my case, the cause was **wrongly written resource definition (serverless.yml)** . By fixing serverless.yml, the problem was solved.

## serverless.yml (having the issue)

I dropped `stateMachines` between `stepFunctions` and `hellostepfunc`, which cause the issue:

```yaml
functions:
  hellofunc:
    handler: handler.hello

stepFunctions:
  hellostepfunc:
    Comment: "A Hello World example of the Amazon States Language using an AWS Lambda Function"
[...]

PR #147, opened 2018-08-29
Refactor function resource arn regex to only allow valid characters

ACSchott526:
- There is a bug with the resource field being empty reported here: https://github.com/horike37/serverless-step-functions/issues/146
- This is due to an issue with the regex not allowing numbers/* character which is common in function arn's.
- I refactored the regex to only allow valid characters for a function arn.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -12,7 +12,7 @@ module.exports = {
       customRolesProvided.push('role' in stateMachineObj);
 
       const stateMachineJson = JSON.stringify(stateMachineObj);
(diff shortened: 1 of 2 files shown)


horike37:
Thanks @ACSchott526 !
LGTM and will cut a patch release soon.

notEvil:
This patch breaks our deployment because we use serverless-pseudo-parameters (for instance #{AWS::Region}). Is there any chance this will be adjusted accordingly?

ACSchott526:
@notEvil https://github.com/horike37/serverless-step-functions/pull/148

ksnyde:
Has this PR made it into the build? I'm still having issues with the `*` wildcard character being used in the **Resource** definition.

PR #148, opened 2018-08-30
Allow pseudo-parameters in function arn regex

ACSchott526:
* Removed some characters in https://github.com/horike37/serverless-step-functions/pull/147 from the regex that I thought weren't valid but they are. So just adding those back.


horike37:
Thanks @ACSchott526 👍 will do a patch release again.

PR #154, opened 2018-10-31
Add dollar sign to function arn regex

ACSchott526:
* Just adding $ as an allowed character in the function arn
* Various users are reporting this issue on https://github.com/horike37/serverless-step-functions/issues/146


horike37:
@ACSchott526 
Thank you for fixing! The code looks good from my perspective.
I want someone who faced the same problem to test it before merging.

PR #161, opened 2019-01-21
Add support for other AWS services

theburningmonk:
- refactored the compile IAM role module to make it easier to support future integrations
- generate appropriate roles for SNS, SQS, DynamoDB, ECS and Batch permissions to mirror the permissions that the AWS Console would generate when asked to create the roles
- consolidate permissions by common actions and then resources, to minimise number of policy statements


horike37:
@theburningmonk 
Thank you for working on this 👍 Works as expected 😄 
Will release next version.

Issue #163, opened 2019-01-23
Cloudwatch alarms


mirkods:
# Feature Proposal

## Description
Hi there,
congratulations for this very good serverless pluging.

I'd like to have the possibility to create a Cloudwatch alarm based on Step Functions executions state in order to subscribe a lambda function to a specific topic.
<a href="https://github.com/ACloudGuru/serverless-plugin-aws-alerts">This plugin</a> works well for lambda functions so I think that something similar applied to Step Functions could be really interesting.

theburningmonk:
@mirkods what are you thinking? I've been thinking of something similar, and this is what I had in mind:

```
stepFunctions:
  stateMachines:
    processPayment:
      name: something-something
      definition:
        ...
      alarms:
        topics:
          ok: ${self:service}-${opt:stage}-alerts-ok
          alarm: ${self:service}-${opt:stage}-alerts-alarm
          insufficientData: ${self:service}-${opt:stage}-alerts-insufficientData
        metrics:
          - executionsTimeOut
          - executionsFailed
          - executionsAborted
          - executionThrottled
```

where the plugin would auto-generate CW twitchy (as in, trip the alarm when there's any error) alarms for the 4 metrics that correspond to failure cases in CloudWatch

![image](https://user-images.githubusercontent.com/546969/51995208-2f699680-24aa-11e9-9394-6258aaffbaff.png)

mirkods:
@theburningmonk this is exactly what I meant and I think that this part can be an important improvement for this module.

theburningmonk:
Awesome, I've been thinking about similar lines as well, as we constantly have to create those alarms either manually or use the ACG plugin, which doesn't offer an opinionated set of defaults that works for step functions.

theburningmonk:
this has been implemented by #167

PR #167, opened 2019-01-31
add support for specifying CloudWatch Alarms

theburningmonk:
Implements the suggestion from #163


horike37:
Thanks again @theburningmonk !
This is Awesome feature! will take a look as soon as possible 👍 😄

PR #169, opened 2019-02-02
Fix Batch/ECS integrations

dpwrussell:
The support for these new integrations was failing to even deploy because "rules" instead of "rule" in the prefix of the Resource was used.

For reference:
https://docs.aws.amazon.com/step-functions/latest/dg/ecs-iam.html
https://docs.aws.amazon.com/step-functions/latest/dg/batch-iam.html


theburningmonk:
thanks! my bad, copy and paste error...

PR #171, opened 2019-02-08
Avoid alarms logs when no alarms are defined

j0k3r:
The way `validateConfig` works will generate some error logs when no alarms are defined at all.

After upgrading to the latest serverless-step-functions we got some logs about alarms but we don't have alarms defined. I think we should not have log about "alarms config is malformed" when no alarms are defined.

Following https://github.com/horike37/serverless-step-functions/pull/167

Sidenote: I've enable `sinon.spy()` instead of `console.log` in `compileIamRole.test.js` to avoid noisy log during tests. I can remove that if you don't agree.

### DIFF
--- a/lib/deploy/stepFunctions/compileAlarms.js
+++ b/lib/deploy/stepFunctions/compileAlarms.js
@@ -74,6 +74,11 @@ function getCloudWatchAlarms(
 }
 
 function validateConfig(serverless, stateMachineName, alarmsObj) {
+  // no alarms defined at all
+  if (!_.isObject(alarmsObj)) {
+    return false;
+  }
+
   if (!_.isObject(alarmsObj) ||
       !_.isObject(alarmsObj.topics) ||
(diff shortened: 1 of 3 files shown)


lib/deploy/stepFunctions/compileAlarms.js
```diff
@@ -74,6 +74,11 @@ function getCloudWatchAlarms(
 }
 
 function validateConfig(serverless, stateMachineName, alarmsObj) {
+  // no alarms defined at all
+  if (!_.isObject(alarmsObj)) {
+    return false;
+  }
+
   if (!_.isObject(alarmsObj) ||
```
theburningmonk: can remove this condition now

j0k3r: Oh yeah I forgot to removed it :+1:

theburningmonk:
This has been deployed as part of `1.12.1`

Issue #166, opened 2019-01-31
Support tagging


noahw3:
# This is a Feature Proposal

## Description

* AWS recently added support for tagging for step functions. It'd be wonderful if this plugin could support tags. I'd imagine it'd be a field alongside role, definition, etc.
* Unfortunately there's no support on the [CreateStepFunction API](https://docs.aws.amazon.com/step-functions/latest/apireference/API_CreateStateMachine.html) right now, meaning that it's not in [CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html) either. It's in separate [TagResource](https://docs.aws.amazon.com/step-functions/latest/apireference/API_TagResource.html) and [UntagResource](https://docs.aws.amazon.com/step-functions/latest/apireference/API_UntagResource.html) calls ([ListTagsByResource](https://docs.aws.amazon.com/step-functions/latest/apireference/API_ListTagsForResource.html) also exists).

theburningmonk:
looks like tags are supported now! yay! 

![image](https://user-images.githubusercontent.com/546969/54378638-b2455a00-46c2-11e9-875f-e48338147107.png)

theburningmonk:
released as part of v1.15.0

PR #181, opened 2019-03-14
Feature/support tagging

theburningmonk:
PR for #166 now that CloudFormation supports tagging for Step Functions

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,9 @@ stepFunctions:
             Resource: arn:aws:lambda:#{AWS::Region}:#{AWS::AccountId}:function:${self:service}-${opt:stage}-hello
             End: true
       dependsOn: CustomIamRole
+      tags:
+        - Key: Team
+          Value: Atlantis
       alarms:
         topics:
           ok: arn:aws:sns:us-east-1:1234567890:NotifyMe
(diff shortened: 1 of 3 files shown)


noahw3:
Looks like this matches the cloudformation template exactly, but it might be nicer syntactically to define the tags as a dictionary (like how lambdas work) and then internally convert it to the array of key/value pairs?

theburningmonk:
@noahw3 I thought about that too, but I have actually seen teams use `:` as delimiter in the key. There's nothing stopping us from still using `:` as delimiter though, e.g. `my:namespace:for:tags:value` => `{ Key: "my:namespace:for:tags", Value: "value" }` but it might look a bit old.

But what if `Value` is an ARN which is also delimited by `:`? It would break our parsing. Open to suggestions as to how we might address that.

noahw3:
If a special character is needed in the actual value, quoting it should be sufficient?

noahw3:
Base serverless project implementation: https://github.com/serverless/serverless/blob/00d129ca86febcd960b17b19818ac413f7bee4d9/lib/plugins/aws/package/compile/functions/index.js#L135-L145

theburningmonk:
@noahw3 ahh, I completely misunderstood you! You meant treat it as an object instead of array, of course, that makes sense :-D

Issue #128, opened 2018-06-27
Allow IAM Role ARN to be passed into CloudWatch Scheduled Event creation


atheiman:
# This is a Feature Proposal

## Description

Users should be able to specify an existing IAM Role ARN in scheduled events rather than having one generated. There is currently an IAM role defined in `iamRoleTemplate` of `compileScheduledEvents.js`:
https://github.com/horike37/serverless-step-functions/blob/673f26b9c4da1ea7b81b55b43b938aa6026fd15f/lib/deploy/events/schedule/compileScheduledEvents.js#L109
This is bolted onto the CloudFormation template generated by `serverless` deploy. Our automation that deploys our Serverless projects does not have permission to deploy these IAM roles so we get this error:

```
An error occurred: myStepFuncScheduleToStepFunctionsRole - 
API: iam:CreateRole User: arn:aws:sts::<accountId>:assumed-role/<myServerlessDeploymentRole> 
is not authorized to perform: iam:CreateRole 
on resource: arn:aws:iam::<accountId>:role/<my-service-name>-myStepFuncScheduleToSt-RANDOMSTRING.
```

Our infrastructure team manages IAM resources, they are able to build us the needed role and we would then specify the Role ARN in the `serverless.yml`:

```yaml
stepFunctions:
  stateMachines:
    myStepFunc:
      role: arn:aws:iam::<accountId>:role/serverless/ScheduledEventRole
      definition:
        # ...
```

## Additional Data

`"serverless": "^1.27.3"`
`"serverless-step-functions": "^1.4.1"`

stevecaldwell77:
I created a PR for this, let me know your thoughts.

theburningmonk:
The change has been published as 1.17.0. This seems like a good change and we can apply this to the cloudwatch event as well. Fancy making another PR? ;-)

PR #186, opened 2019-03-28
feat: allow the specification of a custom IAM role for scheduled events

stevecaldwell77:
Implementation of solution for #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,7 @@ events:
 
 ## Scheduled Events IAM Role
 
-By default, the plugin will create a new IAM role that allows AWS Events to start your state machine. Note that this role is different then the role assumed by the state machine. You can specify your own role instead (it must allow `events.amazonaws.com` to assume it, and it must be able to run `states:StartExecution` on your state machine):
+By default, the plugin will create a new IAM role that allows AWS Events to start your state machine. Note that this role is different than the role assumed by the state machine. You can specify your own role instead (it must allow `events.amazonaws.com` to assume it, and it must be able to run `states:StartExecution` on your state machine):
 
 ```yaml
 events:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -579,6 +579,17 @@ events:
       rate: rate(2 hours)
 ```
 
+## Scheduled Events IAM Role
+
+By default, the plugin will create a new IAM role that allows AWS Events to start your state machine. Note that this role is different then the role assumed by the state machine. You can specify your own role instead (it must allow `events.amazonaws.com` to assume it, and it must be able to run `states:StartExecution` on your state machine):
```
theburningmonk: typo: `different than`

lib/deploy/events/schedule/compileScheduledEvents.js
```diff
@@ -95,11 +95,17 @@ module.exports = {
                     ${InputPath ? `"InputPath": "${InputPath}",` : ''}
                     "Arn": { "Ref": "${stateMachineLogicalId}" },
                     "Id": "${scheduleId}",
-                    "RoleArn": {
-                      "Fn::GetAtt": [
-                        "${scheduleIamRoleLogicalId}",
-                        "Arn"
-                      ]
+                    "RoleArn": ${
```
theburningmonk: instead of inlining everything, how apart pulling it out, e.g. `const roleArn = ...`? not essential, just makes it a bit easier to see the template

theburningmonk:
minor typo and suggestions for styling change, but otherwise looks good to me

lib/deploy/events/schedule/compileScheduledEvents.js
stevecaldwell77: Agreed, done.

README.md
stevecaldwell77: Good catch, thanks.

PR #196, opened 2019-05-08
fix: address some diffs with lambda_proxy

theburningmonk:
1) body is object not string; 
2) ordering of props; 
3) queryStringParameters return {} instead of null; 
4) pathParameters return {} instead of null; 
5) stageVariables return {} instead of null;
6) no requestContext.domainName and requestContext.domainPrefix

cc/ @andrhamm


theburningmonk:
still not addressed: multivalue query string params and headers, not sure how that's done in API Gateway at the moment, don't see the multi-value params in $input.params, pinged some AWS folks to see if they can answer

PR #207, opened 2019-06-04
feat: add support for AuthorizationScopes

### LINKED ISSUE #200: Setting Oauth scope in Api Gateway does not work for step functions
have a step function with http event configured. I am using Cognito authorizer and using oauth scopes. Syntax is the same as with Lambda and API Gateway configuration (which works correctly).

Example of cfg:

```yaml
stepFunctions:
  stateMachines:
    calculate:
            events:
            - http:
                path: v1/calculate
                method: POST
                cors: true
                authorizer:
                  name: authorizer
                  arn: 'arn'
                  identitySouce: method.request.header.Authorization
                  scopes: 
                    - 'scope1'
```

## Additional Data

* ***Serverless Framework Version you're using***:   1.42.3
* ***Provider Error messages***:  deployment works well, authorizer is set but scope is not set
* ***Provider Error messages***:

theburningmonk:
Closes #200

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,6 +497,10 @@ stepFunctions:
               type: COGNITO_USER_POOLS # TOKEN, CUSTOM or COGNITO_USER_POOLS, same as AWS Cloudformation documentation
               authorizerId:
                 Ref: ApiGatewayAuthorizer  # or hard-code Authorizer ID
+              # [Optional] you can also specify the OAuth scopes for Cognito
+              scopes:
+                - scope1
+                ...
 ```
 
 #### LAMBDA_PROXY request template
--- a/lib/deploy/events/apiGateway/methods.js
+++ b/lib/deploy/events/apiGateway/methods.js
@@ -328,6 +328,7 @@ module.exports = {
           Properties: {
             AuthorizationType: http.authorizer.type,
             AuthorizerId: http.authorizer.authorizerId,
+            AuthorizationScopes: http.authorizer.scopes,
           },
         };
       }
@@ -336,10 +337,12 @@ module.exports = {
         .getAuthorizerLogicalId(http.authorizer.name || http.authorizer);
 
       let authorizationType;
+      let authorizationScopes;
       const authorizerArn = http.authorizer.arn;
       if (typeof authorizerArn === 'string'
         && awsArnRegExs.cognitoIdpArnExpr.test(authorizerArn)) {
         authorizationType = 'COGNITO_USER_POOLS';
+        authorizationScopes = http.authorizer.scopes;
       } else {
         authorizationType = 'CUSTOM';
       }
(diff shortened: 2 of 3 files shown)


theburningmonk:
@gregadunbit mind taking a peep at this? especially the test cases

gregadunbit:
@theburningmonk will look into it first thing tomorrow morning.

gregadunbit:
@horike37 can you trigger release for a version with this feature so that I can test it?

theburningmonk:
@gregadunbit done!

gregadunbit:
It doesn't work for me:



@theburningmonk 

This is the syntax which I use for apigw and lambdas:

```yml
events:
        - http:
            path: v1/calculate
            method: POST
            cors: true
            authorizer:
              name: authorizer
              arn: arn:aws:cognito-idp:eu-west-1:xxxxxxxxxx
              identitySouce: method.request.header.Authorization
              scopes: 
                - scope1
```

I tried to add also

```yml
events:
        - http:
            path: v1/calculate
            method: POST
            cors: true
            authorizer:
              name: authorizer
              type: COGNITO_USER_POOLS
              arn: arn:aws:cognito-idp:eu-west-1:732945622201:userpool/eu-west-1_XvVw9TFBh
              identitySouce: method.request.header.Authorization
              scopes: 
                - scope 1
```

Does not work for an existing stack and also for new.

gregadunbit:
I think the problem is in this line:

https://github.com/horike37/serverless-step-functions/blob/07a62777970e95299506f8ad0a3453a677a63736/lib/deploy/events/apiGateway/methods.js#L326

Probably it should be:

if (http.authorizer.type && (http.authorizer.authorizerId || http.authorizer.name) ) {

theburningmonk:
I think that case is for when you use "Custom", but your case should be handled by this line: https://github.com/horike37/serverless-step-functions/blob/07a62777970e95299506f8ad0a3453a677a63736/lib/deploy/events/apiGateway/methods.js#L342-L345

theburningmonk:
but I also don't see what we do with the `identitySouce` field, don't think that's supported either!

theburningmonk:
Also, what's the `name` property supposed to signify?

gregadunbit:
name is set in Authorization field in Method Request in APIGW.

You are right about the code. Is it somehow possible that  awsArnRegExs.
cognitoIdpArnExpr.test(authorizerArn)) does not work well?

gregadunbit:
https://github.com/horike37/serverless-step-functions/blob/07a62777970e95299506f8ad0a3453a677a63736/lib/utils/arnRegularExpressions.js#L4

Should be

cognitoIdpArnExpr: /arn:[a-zA-Z-]*:cognito-idp/,

without ^ ?

theburningmonk:
@gregadunbit no that wouldn't be it, since `^` just asserts the start, although I'm curious why the `lambdaArnExpr` doesn't have it?

just tested it quickly and it should still match
image

theburningmonk:
ok, I think I found the problem, scopes is stripped out before it even reached the `getMethodAuthorization` method!

PR #211, opened 2019-06-04
feat: support lambda:invoke resource type

### LINKED ISSUE #209: Support `Resource: arn:aws:states:::lambda:invoke`
# This is a Feature Proposal

## Description

With the new [Context Object](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html) it is possible to define a state with a resource of `Resource: arn:aws:states:::lambda:invoke`. This is similar to the callback pattern but without the `.waitForTaskToken` suffix. Specifying the function arn in the `Parmeters.FunctionName` in this way (as opposed to in `Resource`) allows you to then also modify the `Parameters.Payload` property, optionally referencing values from the context object using `$$`.

Example:

```
GetExecutionName:
  Type: Task
  Resource: arn:aws:states:::lambda:invoke
  Parameters:
    FunctionName: arn:aws:lambda:us-east-1:1234123412341234:function:my-service-dev-myFunction
    Payload:
      ExecutionName.$: "$$.Execution.Name"
```

It looks like this might be as simple as adding an additional `case` [here](https://github.com/horike37/serverless-step-functions/blob/master/lib/deploy/stepFunctions/compileIamRole.js#L245-L246) like so:
```
case 'arn:aws:states:::lambda:invoke':
case 'arn:aws:states:::lambda:invoke.waitForTaskToken':
    return getLambdaPermissions(state);
```

Similar or dependent issues:
* #202

theburningmonk:
Closes #209

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -242,6 +242,7 @@ function getIamPermissions(serverless, taskStates) {
       case 'arn:aws:states:::ecs:runTask':
         return getEcsPermissions();
 
+      case 'arn:aws:states:::lambda:invoke':
       case 'arn:aws:states:::lambda:invoke.waitForTaskToken':
         return getLambdaPermissions(state);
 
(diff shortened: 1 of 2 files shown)


theburningmonk:
@andrhamm wanna have a look?

andrhamm:
As usual I didn't even have time to finish writing my temporary workaround before you got to it! 🎉 

LGTM

theburningmonk:
@andrhamm to be fair, it's a pretty straight forward change, and you even told me where to change it! Try out the new version, hopefully it works as expected.

andrhamm:
Works!

PR #212, opened 2019-06-06
fix: scopes is now passed through validation

### LINKED ISSUE #200: Setting Oauth scope in Api Gateway does not work for step functions

theburningmonk:
Closes #200

### DIFF
--- a/lib/deploy/events/apiGateway/validate.js
+++ b/lib/deploy/events/apiGateway/validate.js
@@ -151,6 +151,7 @@ module.exports = {
     let identityValidationExpression;
     let claims;
     let authorizerId;
+    let scopes;
 
     if (typeof authorizer === 'string') {
       if (authorizer.toUpperCase() === 'AWS_IAM') {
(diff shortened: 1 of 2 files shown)


theburningmonk:
@gregadunbit this should fix it

theburningmonk:
@gregadunbit this has been deployed in `1.23.1`, care to try again to see if it's working now?

gregadunbit:
@theburningmonk works. Tnx.

PR #220, opened 2019-06-19
Serverless warning fix

jkruse14:
… available

We were getting a bunch of these warnings due to this plugin:

```
Serverless Warning --------------------------------------
 
  A valid service attribute to satisfy the declaration 'self:provider.environment.<var_name>' could not be found.
```

I found the stepFunctions config could be loaded directly in this manner, allowing other variables to not be loaded again.

### DIFF
--- a/lib/yamlParser.js
+++ b/lib/yamlParser.js
@@ -13,35 +13,41 @@ module.exports = {
     const serverlessYmlPath = path.join(servicePath, 'serverless.yml');
     return this.serverless.yamlParser
       .parse(serverlessYmlPath)
-      .then(serverlessFileParam =>
-        this.serverless.variables.populateObject(serverlessFileParam)
-        .then(parsedObject => {
-          this.serverless.service.stepFunctions = {};
-          this.serverless.service.stepFunctions.stateMachines
-              = parsedObject.stepFunctions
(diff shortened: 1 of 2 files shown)


theburningmonk:
Thanks @jkruse14, can you add some tests around the change as well?

jkruse14:
@theburningmonk - I added a test that covers the case if a stepfunctions config is not given. Previous tests cover the case when it is given. Let me know if there's anything else I can do!

theburningmonk:
@jkruse14 I think the code change looks fine, but I wanna understand the error better. Do you have a minimal example that I can use to reproduce the error?

jkruse14:
@theburningmonk - I'll get you one soon 👍

theburningmonk:
@jkruse14 any luck? don't want to keep the PR open for too long or it'll drift more and more

jkruse14:
@theburningmonk - sorry for the delay, life has been a bit chaotic as of late. After some more digging, I noticed that the variables that it was complaining about are environment variables we load through a plugin of our own that run on the same hook as serverless-step-functions. I'm not sure if that *should* cause an issue like this or if there is another "better" or "more correct" solution? That being said, using this update makes it so serverless-step-functions is only concerned about the data it requires.

Issue #72, opened 2017-08-08
Support DescribeExecution action in API Gateway


jakezatecky:
# This is a Feature Proposal

## Description

Thanks for the plugin!

The `serverless-step-functions` plugin provides integration to the API Gateway by adding a hook to invoke the step function by via the `StartExecution` action. When invoked, it returns the following output:

```
{
   "executionArn": "string",
   "startDate": number
}
```

This is not particularly useful if we want to use the step function as an API endpoint to wait for output. AWS supports a [`DescribeExecution`](http://docs.aws.amazon.com/step-functions/latest/apireference/API_DescribeExecution.html) event for step functions that, when the function has been completed, will provide output. Naturally, the `executionArn` has to be provided to invoke this action.

There are other ways to get output (the step function could theoretically call an endpoint to cause some type of web-socket communications), but short polling `DescribeExecution` seems like a common enough use case. It's pretty easy to create an integration to `DescribeExecution` via the API Gateway within the AWS console, but it does not appear to be supported easily in this plugin.

This might necessitate that the `events` configuration be changed a bit to distinguish between `StartExecution` and `DescribeExecution`. E.g.:

``` yaml
events:
   -
    http:
      path: step
      method: get
      cors: true
      action: DescribeExecution
```

Is this at all reasonable?

## Additional Data
* ***Serverless Framework Core Version you're using***: latest
* ***The Plugin Version you're using***: latest
* ***Operating System***: N/A
* ***Stack Trace***: N/A
* ***Provider Error messages***: N/A

horike37:
Thank you for great proposal @jakezatecky :smile:
I agree with the proposal and looks good adding the `action` syntax to 'events' section.

ancil26:
+1 
@jakezatecky This is exactly what we need for our system too, as we're using the front end to hit an API endpoint to retrieve a bunch of objects.

@horike37 can you point me to whereabouts in the source code that will need the necessary changes to allow this?

horike37:
@ancil26 
Great! Thank you for your effort!
Although I don't verify it, but it seems that it would change around [here](https://github.com/horike37/serverless-step-functions/blob/19ab556f83f641804b7ee8de73d8a83dc449009a/lib/deploy/events/apiGateway/methods.js#L64) to specify `DescribeExecution`

Let us know if you have any other questions.

ancil26:
[...]

PR #222, opened 2019-07-04
DescribeExecution as option

shota:
As described in issue #72, DescribeExecution support is very important feature when implementing step functions with API Gateway. However, it seems like I can implement this with very short difference so I wrote this pull request.

Caution: This pull request makes "DescribeExecution" action allowed in the policy.


shota:
Example configuration as below

```
stepFunctions:
  stateMachines:
    command:
      events:
        - http:
            path: command/execute
            method: post
            cors: true
        - http:
            path: command/status
            method: post
            cors: true
            is_describe: true
```

"is_describe: true" makes us happy!

lib/deploy/events/cloudWatchEvent/compileCloudWatchEventEvents.js
```diff
@@ -123,7 +123,8 @@ module.exports = {
                         {
                           "Effect": "Allow",
                           "Action": [
-                            "states:StartExecution"
+                            "states:StartExecution",
+                            "states:DescribeExecution"
```
theburningmonk: you shouldn't need `DescribeExecution` for CloudWatch events, right?

lib/deploy/events/apiGateway/apigateway-to-stepfunctions-assume-role.json
```diff
@@ -22,7 +22,8 @@
             {
               "Effect": "Allow",
               "Action": [
-                "states:StartExecution"
+                "states:StartExecution",
+                "states:DescribeExecution"
```
theburningmonk: maybe take this permission-generation logic out, and put it into the code so that we only give the necessary permissions if the user configures an endpoint to describe execution

theburningmonk:
@shota thanks for putting this together, I have a couple of feedback:

1. I think the original config change proposed in #72 is more extensible, it'll be easier to support other actions like stopping an execution, etc. in the future
```
events:
   -
    http:
      path: status
      method: get
      cors: true
      action: DescribeExecution
```

2. I notice you didn't have to update the integration request template, but `StartExecution` has different input argument to `DescribeExecution`, so how did you test this? e.g. what payload did you post to the describe endpoint to make it work?

3. can you update the relevant section in the README as well to document the new feature?

4. we use `semantic-release` to generate the new version number, so it relies on the commit messages to follow the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) standard. Please update the commit messages so they follow the convention.

shota:
Sounds fantastic, I'll try again later.

shota:
@theburningmonk Thank you for your review, my pull request has some bugs and I fixed them and clean commit messages as you assigned. 

Document was fixed and code was tested with full coverage.

lib/deploy/events/apiGateway/methods.js
```diff
@@ -234,6 +241,12 @@ module.exports = {
         http.request.template
       );
     }
+
+    if (_.has(http, 'action')) {
+      // no template if some action was defined.
+      return {};
```
theburningmonk: Does this mean any user-specified request template won't be used? e.g. if someone wants to use a GET to describe the execution and pass in the execution ARN as a query string param, would they be able to write a custom request template to do that?

shota: No, if user specify http.request.template option, it will be applied. These codes are just avobe `return defaultTemplate`. The default action for StartExecution should not be applied all other actions and should pass users action through.

theburningmonk: right, right, I see. Thanks, that looks good to me.

lib/deploy/events/apiGateway/iamRole.js
```diff
@@ -14,15 +13,58 @@ module.exports = {
...
+        Policies: [
+          {
+            PolicyName: 'apigatewaytostepfunctions',
+            PolicyDocument: {
+              Version: '2012-10-17',
+              Statement: [
+                {
+                  Effect: 'Allow',
+                  Action: iamRoleApiGatewayToStepFunctionsAction,
+                  Resource: '*',
```
theburningmonk: A minor thing, we can change this later in another PR, but we should be able to restrict the resource to just the state machine in question.

theburningmonk:
@shota I didn't realise you had a breaking change commit message in one of the commit messages before I merged. For future reference, it shouldn't count as a breaking change since it doesn't change any of the existing behaviours.

@horike37 so this PR has published a major version where it probably doesn't need to. I can delete the version on NPM and force semantic-release to publish it as `1.26.0` instead. What do you think?

horike37:
@theburningmonk 
Thank you for maintaining it as always 👍 

> I can delete the version on NPM and force semantic-release to publish it as 1.26.0 instead. What do you think?

It should cut a release as `1.26.0`. Please force update 🙏 

 @shota 
Great work 😄 This PR would make many users to be pleased.
ありがとうございましたー！！

shota:
@theburningmonk 
Oh, I didn't recognize these message cause this, I understand now. Thank you for your support!

@horike37 
🎉 Thank you! こちらこそ！

theburningmonk:
@shota no worries, I should have paid more attention when reviewing :-P It's ok, it seems we can't unpublish v2.0.0 so we're just gonna leave it as it is. Thanks again for the contribution!

PR #236, opened 2019-07-29
docs: add explanation for the logical resource names

### LINKED ISSUE #230: How to find or create logicalNameOfResource for CloudFormation intrinsic functions
# This is a Documentation Clarification

## Description

I'm trying to use CloudFormation intrinsic functions to access resource arn's as described in the documentation: https://github.com/horike37/serverless-step-functions#setup. In the example configuration the resource is referenced like `Fn::GetAtt: [HelloLambdaFunction, Arn]`. This is partly due to a lack of knowledge about AWS CloudFormation, but I'm not sure how that string `HelloLambdaFunction` was constructed.

In the AWS documentation this is referred to as the `logicalNameOfResource`: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html. And then there's more detail here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html.

I might be mis-interpreting this, but it looks to me like the `logicalNameOfResource` is defined by the author somewhere in the template. I'm unclear where the functions are assigned this name.

theburningmonk:
Closes #230

@claytron5000 would this explanation suffice?

I'll look at how we can support the local function names - e.g. `hello` - instead of the generated logical resource names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ plugins:
 
 ## Setup
 
-Specifies your statemachine definition using Amazon States Language in a `definition` statement in serverless.yml. You can use CloudFormation intrinsic functions such as `Ref` and `Fn::GetAtt` to reference Lambda functions, SNS topics, SQS queues and DynamoDB tables declared in the same `serverless.yml`.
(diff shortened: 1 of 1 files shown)


claytron5000:
This looks great, thanks!

theburningmonk:
superseded by #237

PR #237, opened 2019-07-30
Feature/support local function name

### LINKED ISSUE #230: How to find or create logicalNameOfResource for CloudFormation intrinsic functions

theburningmonk:
Closes #230

This PR lets you use the local function names (e.g. `hello-world`) instead of logical IDs (e.g. `HelloDashworldLambdaFunction`) which is always a bit confusing to newcomers (and myself!) since it forces you to know how Serverless framework translates those local function names - which makes it a leaky abstraction.

e.g.

```yml
functions:
  hello-world:
    handler: hello-world.js

stepFunctions:
  stateMachines:
    myStateMachine:
      definition:
        StartAt: A
        States:
          A:
            Type: Task
            Resource:
              Fn::GetAtt: [hello-world, Arn] # instead of HelloDashworldLambdaFunction
            End: true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ plugins:
 
 ## Setup
 
-Specifies your statemachine definition using Amazon States Language in a `definition` statement in serverless.yml. You can use CloudFormation intrinsic functions such as `Ref` and `Fn::GetAtt` to reference Lambda functions, SNS topics, SQS queues and DynamoDB tables declared in the same `serverless.yml`.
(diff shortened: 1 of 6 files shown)


theburningmonk:
@claytron5000 hopefully this would make it easier for everybody going forward!

PR #90, opened 2017-12-09
State machine definition validation using asl-validator

ChristopheBougere:
Hi !

I have added a validation command for the state machine definition. All the logic is handled in [asl-validator](https://www.npmjs.com/package/asl-validator) package.

It can help saving a lot of time avoiding failed deployment because of a wrong state machine definition,

You can use it like this: `sls validate-state-machines [--name <stepfunctionname>]`.

I also plan to add it in the `deploy` command, but I'd like it to happen at the beginning (before packaging everything). Let me know what you think about it, and if you have some thoughts about when exactly running validation.

Some output examples:
## Valid definition
```
$ serverless validate-state-machine -vServerless: Validating state machines definition
Serverless: ✓ yourParallelMachine definition is valid
```

## State machine does not exists
```
$ serverless validate-state-machine -n yourParallelMachines
Serverless: Validating state machines definition
 
  Serverless Error ---------------------------------------
 
  stateMachine "yourParallelMachines" doesn't exist in this Service
```

## Invalid defintion (missing StartAt property)
```
$ serverless validate-state-machine
Serverless: Validating state machines definition
Serverless: ✕ yourParallelMachine definition is invalid:
[{"keyword":"required","dataPath":"","schemaPath":"#/required","params":{"missingProperty":"StartAt"},"message":"should have required property 'StartAt'"}]
 
  Serverless Error ---------------------------------------
 
  At least one of your state machine definition is invalid.
```

PR #245, opened 2019-08-16
feat: asl validation

ChristopheBougere:
Replaces #90
This PR adds pre-deployment definition validation using [asl-validator](https://www.npmjs.com/package/asl-validator). The latter [now supports intrinsic CF functions](https://github.com/airware/asl-validator/pull/35), which removes the need for serverless-pseudo-parameters support.

In contrast to #90 which adds a new command (`sls validate stepf`), here we automatically validate the state machine definitions during the deploy step. The main benefit is to avoid uploading and updating CloudFormation stack in order to detect and fix typos, so it can save some time when developing.

The feature is disabled by default, and can be enabled with a boolean:
```yaml
stepFunctions:
  validate: true
```

I think this behaviour is better than the one in #90, but I'm happy to discuss it and revert it.
AFAIK it's working pretty well, but I haven't tested all the new step functions capabilities and associated services so please open issues/PR [here](https://github.com/airware/asl-validator) if you experience issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ This is the Serverless Framework plugin for AWS Step Functions.
      - [CloudWatch Alarms](#cloudwatch-alarms)
      - [CloudWatch Notifications](#cloudwatch-notifications)
      - [Blue-Green deployments](#blue-green-deployment)
+     - [Pre-deployment validation](#pre-deployment-validation)
  - [Current Gotcha](#current-gotcha)
  - [Events](#events)
      - [API Gateway](#api-gateway)
@@ -135,6 +136,7 @@ stepFunctions:
   activities:
     - myTask
     - yourTask
+  validate: true # enable pre-deployment definition validation (disabled by default)
 
 plugins:
   - serverless-step-functions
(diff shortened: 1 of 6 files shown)


theburningmonk:
I agree that the validation should be done automatically. Is there any reason why it shouldn't be enabled by default?

ChristopheBougere:
> Is there any reason why it shouldn't be enabled by default?

No special reason, maybe we can start like this as a "beta" and turn it on by default in a month or so?

theburningmonk:
@ChristopheBougere ok, no worries, can leave it as is so we're backward compatible

@horike37 this looks good from my perspective, can you have a look?

horike37:
@ChristopheBougere @theburningmonk 
Thank you for re-working on this 👍 
The pre-deployment validation looks good but why don't you add `sls validate stepf` at this time?
Imo, the command would make writing the state machine definitions easy since it can do manual validation while writing it.

What do you think?

ChristopheBougere:
@horike37 I agree that it would be nice to have both possibilities.
At the same time, I think it is better to keep it simple in the code and have the same behavior in both cases.
I tried by adding this, which triggers the same functions than in `package:compileFunctions`: https://github.com/horike37/serverless-step-functions/commit/e2c9df9bdcc0d1745f623ec048b108adf05a5df0
But I am having this error:
```
TypeError: Cannot read property 'Resources' of undefined
    at ServerlessStepFunctions.translateLocalFunctionNames (/Users/christophe/my-service/node_modules/serverless-step-functions/lib/utils/aws.js:11:85)
```
Meaning that the variable `this.serverless.service.provider.compiledCloudFormationTemplate` hasn't been populated yet. I'm affraid that some package/deploy steps are missing to make the function name translations.
Maybe you have an idea?

horike37:
@ChristopheBougere 
Ah, ok. I re-thought that while testing this feature and `sls validate stepf` would be unnecessary. Because running the validation at the same time when running `sls package`.

horike37:
@ChristopheBougere 
Could you fix the conflicts? Afterwards, we can merge it.

ChristopheBougere:
@horike37 done 🙂

PR #247, opened 2019-08-18
Feature/support_exact_version

### LINKED ISSUE #240: Allow reference against specific version of function
# This is a Feature Proposal

## Description

To make it easy to implement blue-green deployments for step functions (hence ensuring that running executions are not affected by deployments), we can add a configuration option to reference the current version of functions.

e.g.

```
stepFunctions:
  stateMachines:
    myStateMachine:
      name: blah
      useExactVersion: true # default to false for backward compatibility
      definition:
        StartAt: Hello
        States:
          Hello:
            Type: Task
            Resource:
              Fn::GetAtt: [hello, Arn] # we'll replace this
            End: true
```

When `useExactVersion` is `true`, we can swap out the function ARN with the ARN of the version to be deployed.

We'll also need to adjust the IAM roles that are generated to allow `InvokeFunction` permission on all versions (e.g. `arn:aws:lambda:us-east-1:123456789012:function:hello:*`) as well. Otherwise, after a deploy, the existing executions would lose the ability to invoke the previous version of the functions.

More info on how I managed to do this with alias instead of versions: https://theburningmonk.com/2019/08/how-to-do-blue-green-deployment-for-step-functions/

theburningmonk:
Closes #240 

To enable this feature, I had to make two changes:

1. to add `lambda:InvokeFunction` permissions to `<function arn>:*` always, since state machines share the same IAM role (unfortunately). This is to cater for the fact that if you deploy with `useExactVersion: false` then deploy again with `useExactVersion: true` then the shared IAM role would need to execute against both `arn:aws:...:function-name` as well as `arn:aws:...:function-name:version`.

2. add a configuration for `useExactVersion` and when it's enabled, iterate through the captured function ARNs and swap them out for the corresponding Lambda version.

Besides these, I added a schema definition for the state machine object itself so it's defined in one place and we can remove quite a few if conditions from `compileStateMachine.js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ This is the Serverless Framework plugin for AWS Step Functions.
      - [Depending on another logical id](#depending-on-another-logical-id)
      - [CloudWatch Alarms](#cloudwatch-alarms)
      - [CloudWatch Notifications](#cloudwatch-notifications)
+     - [Blue-Green deployments](#blue-green-deployment)
  - [Current Gotcha](#current-gotcha)
  - [Events](#events)
      - [API Gateway](#api-gateway)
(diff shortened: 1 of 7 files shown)


horike37:
@theburningmonk 
The code looks good! but the doc lacks. can you also add it?

theburningmonk:
@horike37 yup, will do

theburningmonk:
@horike37 done

PR #248, opened 2019-08-20
docs: add example for separate files

NoxHarmonium:
Our serverless.yml files were getting a bit unwieldy with too many step functions so I thought this change might help slim them down by allowing us to extract each step function definition to a file.

This mimics the existing behaviour of the `resources:` section in serverless.yml

See: https://github.com/serverless/serverless/pull/2842

I've tested the code by deploying the step functions as in-line definitions and as external file and checking that there were no differences in the CloudFormation stack.

Let me know if there is anything else I need to do.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1020,6 +1020,78 @@ plugins:
   - serverless-step-functions
 ```
 
+### How to split up state machines into files
+
+When you have a large serverless project with lots of state machines
+your serverless.yml file can grow to a point where it is unmaintainable.
+
+You can split step functions into external files and import them
+into your serverless.yml file.
+
+There are two ways you can do this:
+
+#### Single external file
+
+You can define the entire `stateMachines` block in a separate file
+and import it in its entirety.
+
+includes/state-machines.yml:
+
+```yml
+stateMachines:
+  hellostepfunc1:
+    name: myStateMachine1
+    definition:
+      <your definition>
+  hellostepfunc2:
+    name: myStateMachine2
+    definition:
+      <your definition>
+```
+
+serverless.yml:
+
+```yml
+stepFunctions:
+  ${file(includes/state-machines.yml)}
+
+plugins:
+  - serverless-step-functions
+```
+
+#### Separate Files
+
+You can split up the `stateMachines` block into separate files.
+
+includes/state-machine-1.yml:
+
+```yml
+name: myStateMachine1
+definition:
+  <your definition>
+```
+
+includes/state-machine-2.yml:
+
+```yml
+name: myStateMachine2
+definition:
+  <your definition>
+```
+
+serverless.yml:
+
+```yml
+stepFunctions:
+  hellostepfunc1:
+    ${file(includes/state-machine-1.yml)}
+  hellostepfunc2:
+    ${file(includes/state-machine-2.yml)}
+
+plugins:
+  - serverless-step-functions
+```
+
 ## Sample statemachines setting in serverless.yml
 
 ### Wait State


theburningmonk:
@NoxHarmonium thanks for the contribution, but I'm not convinced this is the way to go. If you didn't turn `stepFunctions.stateMachines` into an array, then there's no need for any change, right?

e.g.
```
stepFunctions:
  stateMachines:
    ${file(includes/my-state-machine.yml)}
```

or if you don't include the state machine itself in the external YMLs, e.g.

```
name: myStateMachine
definition:
  <your definition>
```
then you can do this:
```
stepFunctions:
  stateMachines:
    myStateMachine:
      ${file(includes/my-state-machine.yml)}
    myOtherStateMachine:
      ${file(includes/my-other-state-machine.yml)}
```
which still achieves the goal of breaking things up into several files.

I don't think we should support `stepFunctions.stateMachines` being both an array as well as an object, especially since we need the keys (e.g. `myStateMachine` and `myOtherStateMachine`) to be unique as we, by default, use them to generate the logical IDs. This also aligns with the fact that in the `serverless.yml`, `functions` is an object, not an array.

@horike37 what do you think?

NoxHarmonium:
@theburningmonk I don't think that the first example works for multiple files:

```
stepFunctions:
  stateMachines:
    ${file(includes/my-state-machine-a.yml)}
    ${file(includes/my-state-machine-b.yml)}
```

I get the following error:

> Trying to populate non string value into a string for variable ${file(includes/step-functions/scribblr-customer-order-workflow.yml)}. Please make sure the value of the property is a string.

However, your second example does work which kind of makes my change redundant. It might be better if I just put up a PR to document how to reference files that way instead of this change.

The only argument I have in favour of my change is that it makes the behaviour consistent with the Serverless `resources:` section.

theburningmonk:
@NoxHarmonium the first example should work when you have only one file and contains all of your state machines including their keys (like you had in the README).

e.g. sfn.yml
```yml
myStateMachine:
  definition:
     ...

myOtherStateMachine:
  definition:
    ...
```
serverless.yml
```
stepFunctions:
  stateMachines:
    ${file(sfn.yml)}
```
but you can't stack the file inports on top of each other.

What do you mean it's more consistent with the `resources` section? That is also a map, not an array.

NoxHarmonium:
> What do you mean it's more consistent with the resources section? That is also a map, not an array.

It can either be a map or an array.

Specifying an array is useful to split up groups of resources into files.

An example of this is:

https://github.com/AnomalyInnovations/serverless-stack-demo-api/blob/master/serverless.yml#L128-L137

The part of the Serverless library code that I based this change on is:

See: https://github.com/dougmoscrop/serverless/blob/master/lib/classes/Service.js#L88

I copied how that was implemented to try and make it consistent and reduce confusion.

theburningmonk:
ah, I see! Thank you for that, learnt something new there :-)

I'm still not convinced that we should support two ways of doing the same thing, but seeing as there's a strong precedent in the core already I do feel less strongly about my position. I see the reason why they support that for resources since you can end up with quite a lot of different resources. Not sure if that'd be the case for step functions.

@horike37 do you have any thoughts on this?

@NoxHarmonium given the option available already, how much value do you think this change would add? how many step functions do you have in your `serverless.yml`? If the solution I suggested would suffice, would you happy to change the PR to just update the docs and we leave this as something we can do if the use case comes up?

NoxHarmonium:
> how much value do you think this change would add? how many step functions do you have in your serverless.yml

FWIW We have 11 step functions which add up to 1832 LOC which is what prompted the change 😅

> If the solution I suggested would suffice, would you happy to change the PR to just update the docs and we leave this as something we can do if the use case comes up?

This sounds good, we already switched over to using the method you suggested to split the step functions into separate files so we didn't have to use our fork as a dependency and it works well.

I've pushed up a change to the README.

Thanks!

horike37:
@theburningmonk 
Ah, Sorry. I missed the comments.
Agree with your thought. It would better to add only the way to README for now. 

@NoxHarmonium 
Thank you for adding the nice example 💯

Issue #253, opened 2019-09-07
Unable to use cross-stack variable for state machine definition's QueueUrl parameter


jcortega:
# This is a Bug Report 

## Description
When using exported variable from another stack/service for queue url (ie Fn::ImportValue), plugin errors out with the message below.

```
Serverless: Packaging service...
Unable to parse SQS queue url [[object Object]]
Unable to parse SQS queue url [[object Object]]
Serverless: Uploading CloudFormation file to S3...
Serverless: Uploading artifacts...
Serverless: Validating template...
Serverless: Updating Stack...
Serverless: Checking Stack update progress...
....
Serverless: Operation failed!
Serverless: 

  Serverless Error ---------------------------------------
  An error occurred: IamRoleStateMachineExecution - Policy statement must contain resources. (Service: AmazonIdentityManagement; Status Code: 400; Error Code: MalformedPolicyDocument; Request ID: cfeb18ea-d13a-11e9-802b-b9c99487764f).
```
Error occurs when the plugin tries to generate IAM policy statement given the QueueUrl parameter. However, since Fn::ImportValue requires fetching current state from Cloudformation, the URL is not readily available for the plugin to parse.

## Additional Data
* ***Serverless Framework Core Version you're using***: 1.46.0
* ***The Plugin Version you're using***: 2.6.0
* ***Operating System***: amazon linux

## Step function definition
```stepFunctions:
  stateMachines:
    updateCMI:
      events:
        - http:
            path: updatecmi
            method: post
            cors: true
            authorizer:
              type: CUSTOM
              authorizerId: {"Fn::ImportValue": "authorizer-${file(../../config/provider.yml):stage}-id"}
      definition:
        StartAt: updateInfra
        States:
          updateInfra:
            Type: Task
            TimeoutSeconds: 300
            Parameters:
              QueueUrl: 
                'Fn::ImportValue': infradeployer-${self:custom.stage}-queue-url
              MessageAttributes:
                IntegrationID:
                  DataType: String
                  StringValue.$: $.integration_id
                OrganizationID:
                  DataType: String
                  StringValue.$: $.organization_id
                Action:
                  DataType: String
                  StringValue.$: $.action
                TaskToken:
                  DataType: String
                  StringValue.$: $$.Task.Token
              MessageBody: "{}"
            Resource: arn:aws:states:::sqs:sendMessage.waitForTaskToken
            Next: choiceStateAction
            ResultPath: $.taskresult
            Retry:
[...]

PR #261, opened 2019-10-01
Fix SQS QueueUrl transformation to ARN when intrinsic function being passed in

FilipPyrek:
Should solve: #253 

I got an idea that maybe it would be nice if we could set QueueArn in State Machine definition like this:
```yaml
some-step:
  Type: Task
  Resource: arn:aws:states:::sqs:sendMessage.waitForTaskToken
  Parameters:
    QueueUrl: { "Fn::ImportValue": "some-queue-url" }
    QueueArn: { "Fn::ImportValue": "some-queue-arn" }
    Message: "42"
  End: true
```
But I wasn't sure how this would fit to architecture of whole project.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -35,12 +35,18 @@ function sqsQueueUrlToArn(serverless, queueUrl) {
     const accountId = match[2];
     const queueName = match[3];
     return `arn:aws:sqs:${region}:${accountId}:${queueName}`;
-  } if (isIntrinsic(queueUrl) && queueUrl.Ref) {
-    // most likely we'll see a { Ref: LogicalId }, which we need to map to
-    // { Fn::GetAtt: [ LogicalId, Arn ] } to get the ARN
-    return {
-      'Fn::GetAtt': [queueUrl.Ref, 'Arn'],
-    };
+  }
+  if (isIntrinsic(queueUrl)) {
+    if (queueUrl.Ref) {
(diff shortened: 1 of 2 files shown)


theburningmonk:
Doesn't look like Step Functions supports QueueARN for the SQS integration. If you did that, you'd have to translate the ARN into a queue URL. I'm not seeing the value of that, got any use case in mind?

lib/deploy/stepFunctions/compileIamRole.js
```diff
@@ -35,12 +35,18 @@ function sqsQueueUrlToArn(serverless, queueUrl) {
...
+    if (queueUrl.Ref) {
+      // most likely we'll see a { Ref: LogicalId }, which we need to map to
+      // { Fn::GetAtt: [ LogicalId, Arn ] } to get the ARN
+      return {
+        'Fn::GetAtt': [queueUrl.Ref, 'Arn'],
+      };
+    }
+    // in case of for example { Fn::ImportValue: sharedValueToImport }
+    // we need to use "*" as ARN
+    return '*';
```
theburningmonk: ah, now I see why you want to add both QueueUrl as well as QueueArn, so you can ref the Arn in the IAM instead of QueueUrl. It'd be doable for sure, but it'd be a bit odd though, and hard to explain to people. Let's leave it as it is for now.

FilipPyrek: Yeah, that's what I meant. I agree with you. :slightly_smiling_face: :+1:

PR #266, opened 2019-10-10
docs: add map state example

hideokamoto:
Example YAML to launch a state machine using Map state.


theburningmonk:
@hideokamoto thank you for the contribution, totally overlooked this part of the documentation when we added support for Map states! :-D

PR #254, opened 2019-09-12
feat: add all authorizer context properties to lambda_proxy template

djiworks:
It seems in the `LAMBDA_PROXY_REQUEST_TEMPLATE`, all the variables `$context.authorizer.*` are omitted.

Looping on the `$context.authorizer.*` allow you to get back the data you pass in a policy generated by a custom authorizer.

Source : https://docs.aws.amazon.com/en_en/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html and https://docs.aws.amazon.com/en_en/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html

PR #278, opened 2019-11-23
fix invalid lambda_proxy template

orzarchi:
It seems like #254 broke the lambda proxy template by introducing an extra comma at the end of a loop, causing an invalid json structure.
This is a simple fix - moving the loop so it's not the end of the json structure


orzarchi:
Hi everyone, any chance for a merge? This is a really simple change, and without it the lambda proxy is currently unusable.

PR #235, opened 2019-07-25
Fix VTL simple quote escaping

thibaultdalban:
Follow up PR of #141 Closes #140

Thanks to @marcelalburg for the initial work done and to @lendesk-pawel-cisek for his advise on how to test it.

PR #279, opened 2019-11-23
Fix serialization errors in lambda_proxy template when using single quotes in payload

orzarchi:
Applying the single quote fix from #235 to lambda proxy template


theburningmonk:
@orzarchi could you add some tests for this as well?

orzarchi:
Added test

orzarchi:
@theburningmonk Hey, any chance for a merge? I've added a test.

PR #300, opened 2020-01-14
Update to README

SamphireReeve:
Add more detail around IAM roles for serverless step functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -981,7 +981,7 @@ Run `sls deploy`, the defined Stepfunctions are deployed.
 
 ## IAM Role
 
-The IAM roles required to run Statemachine are automatically generated for the state machines lambda, with the policy name of `StatesExecutionPolicy-<environment>`. This is given the default permissions of allowing the lambda InvokeFunction. You can also specify a custom ARN directly to the step functions lambda.
+The IAM roles required to run Statemachine are automatically generated for each state machine in the `serverless.yml`, with the IAM role name of `StatesExecutionPolicy-<environment>`. These roles are tailored to the services that the state machine integrates with, for example with Lambda the `InvokeFunction` is applied. You can also specify a custom ARN directly to the step functions lambda.
 
 Here's an example:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -981,7 +981,7 @@ Run `sls deploy`, the defined Stepfunctions are deployed.
 
 ## IAM Role
 
-The IAM roles required to run Statemachine are automatically generated. It is also possible to specify ARN directly.
+The IAM roles required to run Statemachine are automatically generated for the state machines lambda, with the policy name of `StatesExecutionPolicy-<environment>`. This is given the default permissions of allowing the lambda InvokeFunction. You can also specify a custom ARN directly to the step functions lambda.
```
theburningmonk: Please rephrase "state machines lambda" to just "state machine" and add a mention that an IAM role is generated for each state machine in the `serverless.yml` and the permissions are tailored to the services that state machine integrates with - Lambda, SNS, SQS, etc (it's not just Lambda `InvokeFunction`).

Also, when you said `policy name` did you mean `IAM role name`? That format has changed since 2.14.0 when we moved to 1 IAM per state machine.

README.md
```diff
@@ -993,7 +993,9 @@ stepFunctions:
       definition:
 ```
 
-It is also possible to use the [CloudFormation intrinsic functions](https://docs.aws.amazon.com/en_en/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) to reference resources from elsewhere:
+It is also possible to use the [CloudFormation intrinsic functions](https://docs.aws.amazon.com/en_en/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) to reference resources from elsewhere. This allows for an IAM Role to be created and applied to the step function all within the serverless file.
```
theburningmonk: please rephrase `applied to the step function all within the serverless file` to `applied to the state machines all within the serverless file`

README.md
```diff
@@ -1009,7 +1011,30 @@ resources:
...
+              - sts:AssumeRole
+        Policies:
+          - PolicyName: statePolicy
+            PolicyDocument:
+              Version: version
+              Statement:
+                - Effect: Allow
+                  Action:
+                    - lambda:InvokeFunction
+                  Resource: "*"
```
theburningmonk: please don't use `*` in the example, people would copy it into their project without realising it's bad practice

README.md
```diff
@@ -993,7 +993,9 @@ stepFunctions:
       definition:
 ```
 
-It is also possible to use the [CloudFormation intrinsic functions](https://docs.aws.amazon.com/en_en/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) to reference resources from elsewhere:
+It is also possible to use the [CloudFormation intrinsic functions](https://docs.aws.amazon.com/en_en/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) to reference resources from elsewhere. This allows for an IAM Role to be created and applied to the step function all within the serverless file. 
+
+The below example shows the policy needed if your step function needs the ability to send a message to an sqs queue. To apply the role either the RoleName can be used as a reference in the state machine, or the role ARN can be used like in the example above. It is important to note that if lambda roles are stored in a folder, this must be specified on the `Path` property on the new role.
```
theburningmonk: I don't understand what you mean by `if lambda roles are stored in a folder`, are you referring to the path you want to assign to the state machine role?

README.md
```diff
@@ -1009,7 +1011,30 @@ resources:
     StateMachineRole:
       Type: AWS::IAM::Role
       Properties:
-        ...
+        RoleName: RoleName
+        Path: /path_to_lambda_roles/
```
theburningmonk: maybe `path_of_state_machine_roles`?

README.md
SamphireReeve: Yes should have been IAM role name, sorry for the confusion.

README.md
SamphireReeve: Yes I think I meant the path that the role should be under - rephrased in the update.

PR #311, opened 2020-02-21
feat:  human readable errors in state schema validation

### LINKED ISSUE #310: Human readable validation errors
# This is a Feature Proposal

## Description

* What is the use case that should be solved.

I want to read easily validation errors that comes from `asl-validator` so I can quickly fix them without having to decode JSON.

Here the output I get in case of error:

```
ServerlessError: ✕ State machine "machine1" definition is invalid:
  [{"keyword":"additionalProperties","dataPath":".States['Hello World']","schemaPath":"#/additionalProperties","params":{"additionalProperty":"Resource"},"message":"should NOT have additional properties"}, ...]
```

Here's an example of what it could like once formatted:
```
ServerlessError: ✕ State machine "machine1" definition is invalid:
  - ".States['Hello World']" should NOT have additional property "Resource" [additionalProperties]
  - ...
```

aslafy-z:
Leverage https://github.com/airware/asl-validator/pull/46 feature to show humanized version of errors to user (instead of raw JSON).

Depends on https://github.com/airware/asl-validator/pull/46
Fixes #310

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,12 +44,12 @@
   },
   "dependencies": {
     "@hapi/joi": "^15.0.2",
-    "asl-validator": "^1.5.0",
+    "asl-validator": "^1.7.0",
     "aws-sdk": "^2.282.1",
     "bluebird": "^3.4.0",
     "chalk": "^1.1.1",
     "lodash": "^4.17.11",
-    "serverless": "^1.64.1"
+    "serverless": "^1.72.0"
   },
   "husky": {
     "hooks": {
--- a/lib/deploy/stepFunctions/compileStateMachines.js
+++ b/lib/deploy/stepFunctions/compileStateMachines.js
@@ -114,13 +114,13 @@ module.exports = {
 
         if (stateMachineObj.definition) {
           if (this.serverless.service.stepFunctions.validate) {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "@hapi/joi": "^15.0.2",
-    "asl-validator": "^1.5.0",
+    "asl-validator": "github:aslafy-z/asl-validator#patch-1",
```
horike37: Why did you use your patch one? It would had better to be merged it to asl-validator. Is it possible?

aslafy-z: I'm waiting for my other PR to be merged and released before I can switch back the the real semver. Added a WIP tag in the title.

horike37: Thanks! Agreed with that 👍

theburningmonk:
@aslafy-z looks like your change has been merged into the asl-validator as of v1.7.0, wanna come back and finish off this PR?

aslafy-z:
Should be good by now! @theburningmonk 
Can you have a try from your side? I have no ready stack to do that in here. Thank you!

theburningmonk:
@aslafy-z looks good, just tested it out on a stack

PR #324, opened 2020-04-07
feat: not output cloudformation outputs section

CM-Kajiwara:
This option suppresses output to the Outputs section of Cloudformation.

https://dev.classmethod.jp/articles/serverless-framework-stepfunctions-many-statemachine-workaround/

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,7 +353,7 @@ stepFunctions:
 
 ### Disable Output Cloudformation Outputs section
 
-Disables output in Outputs section.
+Disables output in the CloudFormation Outputs section.  if you define many state-machines in serverless.yml, there is a possibility of reaching out a CloudFormation limits in which the maximum number of outputs is 60. If you define `noOutput: true`, you can prevent automatically output by this plugin.
 
 ```yaml
 stepFUnctions:
--- a/README.md
+++ b/README.md
@@ -351,6 +351,15 @@ stepFunctions:
   validate: true
 ```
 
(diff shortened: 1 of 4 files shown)


lib/deploy/stepFunctions/compileStateMachines.test.js
```diff
@@ -1390,4 +1390,24 @@ describe('#compileStateMachines', () => {
...
+          name: 'stateMachineBeta2',
+          definition: 'definition2',
+          role: 'arn:aws:role2',
+        },
+      },
+      noOutput: true,
+    };
+    serverlessStepFunctions.compileStateMachines();
+    expect(Object.keys(serverlessStepFunctions.serverless.service.provider
+      .compiledCloudFormationTemplate.Outputs).length).to.equal(0);
```
horike37, line 1411: Could you add a test case which `noOutput: false` is explicitly set as well?

README.md
```diff
@@ -351,6 +351,15 @@ stepFunctions:
   validate: true
 ```
 
+### Disable Output Cloudformation Outputs section
+
+Disables output in Outputs section.
```
horike37, line 356: It may lack the explanation in what kind of situation would it be possible to use this.
What about updating something like this?

```
Disables output in the CloudFormation Outputs section.  if you define many state-machines in serverless.yml, there is a possibility of reaching out a CloudFormation limits in which the maximum number of outputs is 60. If you define `noOutput: true`, you can prevent automatically output by this plugin.
```

PR #326, opened 2020-04-15
Trigger error when 'definition' is missing

benja-M-1:
Hello 👋

Today I had a typo in my configuration, I wrote "definitions" instead of "definition"...
Serverless was telling me "TypeError: Cannot read property 'States' of undefined" after some time when deploying.
Without opening the code I couldn't understand what happend. Thus I added this error to help end user to debug faster :)


benja-M-1:
I fixed the build :)

PR #333, opened 2020-05-09
Add IAM Role generation support for resource "startExecution.sync:2"

daaru00:
Hi,

I added a support for IAM Role generation when using `arn:aws:states:::states:startExecution.sync:2`. As [AWS documentation](https://docs.aws.amazon.com/step-functions/latest/dg/connect-stepfunctions.html) describe
state output specification

the `:2` version can be used to have JSON output of nested step function instead of encoded string.


JeremyDOwens:
I hope this gets approved quickly. I was just about to add this today!

PR #343, opened 2020-06-10
feat: retain config

DmytroKucheryavenko:
add ability to set DeletionPolicy to Retain for State Machine


theburningmonk:
@DmytroKucheryavenko can you add a mention to the `retain` option in the `README` as well? otherwise, the changes look good to me

Issue #347, opened 2020-06-12
Move aws-sdk to devDependencies?


nampas:
# This is a Feature Proposal

## Description

When serverless packages up a node project, it removes all devDependencies from the node_moduels directory. This is nice because dev dependencies can contain lots of heavyweight modules (e.g. TypeScript) that we don't want to bother shipping off to AWS.
Unfortunately, right now, serverless-step-functions brings in aws-sdk as a package dependency, which means that all ~50mb of that monster are package up and sent to AWS on each deploy. AWS includes aws-sdk in its nodejs runtime by default, so doing this is not necessary.

Can aws-sdk be made a dev dependency? That way any tests and local invocations of serverless will work fine, but will still allow serverless to remove aws-sdk automatically when packaging.

nampas:
On that note, it'd also be worth removing serverless as a dependency entirely. The expectation is that serverless has already been installed globally

theburningmonk:
@nampas in your project you should install plugins like `serverless-step-functions` as dev dependencies, so that when you package your functions none of it would get into your artefact.

theburningmonk:
for example, [this project](https://github.com/theburningmonk/lambda-step-functions-demo)'s artefact is about 50kb, and it's only this big because of the `package-lock.json`

theburningmonk:
@horike37 looking through the code, we don't actually require the `aws-sdk` anywhere, any idea why it's a dependency in the project at all?

horike37:
@theburningmonk 
That was my careless mistake when the first release 😅 Should move to devDependencies.

theburningmonk:
@horike37 we don't actually need the `aws-sdk`, right? can we remove it altogether?

horike37:
Yup, right 🙆‍♂️

nampas:
Hey thanks for the responses y'all. Happy to create a PR removing aws-sdk if you'd like.

PR #348, opened 2020-06-16
fix: #347 Remove aws-sdk dependency

nampas:
Addresses ticket #347 

Should this be a `feat`?


theburningmonk:
@nampas maybe a `feat` is more appropriate, but it probably doesn't matter in this case.

Issue #355, opened 2020-08-05
Configuration error at root: unrecognized property 'stepFunctions'


TheEdgeOfCat:
# This is a Bug Report

## Description

I'm getting a serverless configuration validation error when deploying with the step function plugin:

```
$ npx serverless deploy
Serverless: Configuration error at root: unrecognized property 'stepFunctions'
Serverless: If you prefer to not continue when error in configuration is approached, ensure `configValidationMode: error` in your config
Serverless: If errors are influenced by an external plugin, enquiry at plugin repository so schema extensions are added (https://www.serverless.com/framework/docs/providers/aws/guide/plugins#extending-validation-schema)
Serverless: If errors seem invalid, please report at https://github.com/serverless/serverless/issues/new?template=bug_report.md
Serverless: If you find this functionality problematic, you may turn it off with `configValidationMode: off` setting
```

## Additional Data
* ***Serverless Framework Core Version you're using***: 1.78.1
* ***The Plugin Version you're using***: 2.21.1
* ***Operating System***: Linux (node v14.7.0)

theburningmonk:
@TheEdgeOfCat ooh, this is interesting, so the serverless framework started to do schema validation, that's new!

TheEdgeOfCat:
Yes, Starting with [1.78.0](https://github.com/serverless/serverless/blob/master/CHANGELOG.md) :)

Maybe I should change this to a feature request instead? I'm not sure how you'd classify it. I could take a look into adding the schema, but probably not before Friday.

theburningmonk:
If you could take a look and pull a PR together, that'd be great. Looks like there's a workaround for now anyway, by turning validation off.

TheEdgeOfCat:
The validation is not enforced at this point anyway, so nothing is really broken. It's just an annoying red error that shows up as the first thing when you run `serverless deploy`

PR #357, opened 2020-08-07
Add basic schema validation

TheEdgeOfRage:
For now I have added only the most basic validation for the top level structure to silence the warning and allow people to use the stricter schema validation.

Closes: #355


theburningmonk:
Thanks @TheEdgeOfRage

PR #364, opened 2020-08-26
feat(codebuild): add IAM policy statement generation for codebuild

eoinsha:
- Add CodeBuild and EventBridge permissions required for codebuild Tasks


theburningmonk (approved):
LGTM

Issue #368, opened 2020-09-04
Warning message when deploying because of dependency on `serverless`


saadtazi:
# This is a minor annoyance

## Description

With serverless installed globally, I updated it using `serverless i -g serverless` to 1.82.0, I now get the following message on a project (with multiple serverless.wml files) using `sls deploy --stage prod` for example (first output of the command):

```
Serverless: Deprecation warning: Local installation of Serverless detected. Starting with next major version, CLI will run it instead of globally installed version.
            Set "enableLocalInstallationFallback" to "true" to switch to new behavior now, set to "false" to keep current behavior and hide this message
            More Info: https://www.serverless.com/framework/docs/deprecations/#LOCAL_INSTALLATION_FALLBACK
```

Once I remove this package and install a serverless.yml that doesn't need this plugin, No warning message.

Nothing is broken, just a little bit annoying...

Maybe:
* remove the hard dependency on `serverless` package [here](https://github.com/serverless-operations/serverless-step-functions/blob/master/package.json#L51), if possible and maybe switch it to a peerDependency (if that's still a thing...)
* is there a dependency on a specific version of version that would prevent from removing the dependency from package.json?

Similar or dependent issues:
* didn't find any

## Additional Data
* ***Serverless Framework Core Version you're using***: 1.82.0
* ***The Plugin Version you're using***: 2.17.2
* ***Operating System***: macOS
* ***Stack Trace***: n/a
* ***Provider Error messages***: n/a

Thank you!

saadtazi:
some clarifications: I have the following structure:

```
/project/
    package.json <-- where serverless-step-functions and other plugins are in `dependencies`
    api/
        service1/
            serverless.yml
        service2/
            serverless.yml
        ...
    pipeline/
        p1/
            serverless.yml <-- where serverless-step-function is used
    ...
```
Installing serverless in `api/package.json` will not work if I do `cd service2/ && sls deploy ...(`sls: command not found`). I would have to install serverless under `service1/`, `service2/`... which doesn't seem to be very efficient.

paulbailey:
I'm also seeing the same issue, and this is the only Serverless plugin that I use that declares a dependency in this way.

lkhari:
Sorry i forgot to auto close this within the PR that was done. But this should be fixed now.

PR #371, opened 2020-09-13
feat: moved authorizers out into code

lkhari:
Removed the Dependency on the serverless library resolving issue #368 

It's pretty much copied from the library which seems wrong, but the rest of the deploy/events/apiGateway code are pretty much the same.


theburningmonk:
Thanks @lkhari yeah, not crazy about copying code from the serverless framework but they haven't exactly made it easy 🤷

Issue #373, opened 2020-09-15
feat: add support for X-ray tracing


bryantbiggs:
# This is a Feature Proposal

## Description

For feature proposals:
- It would be great to add support for enabling x-ray tracing to the plugin natively - new feature announced just yesterday https://aws.amazon.com/blogs/compute/introducing-aws-x-ray-new-integration-with-aws-step-functions/

hassaku63:
It is able to enabling X-Ray by added tracingConfig parameter #374 , but I feel this is not enough as a feature to supports X-Ray.

In order to fully supports X-Ray, I think better to it need to adding a feature generate IAM Policy for StateMachine to use X-Ray  automatically.

theburningmonk:
@hassaku63 fancy putting another PR to add the IAM permissions?

hassaku63:
@theburningmonk yes, i'd like to submit another PR

bryantbiggs:
thanks @hassaku63 and @theburningmonk for the quick response - I think those two PRs close out this issue :+1:

PR #375, opened 2020-09-17
fix: added iam permission for x-ray

hassaku63:
#373 

It generates IAM permission for StateMachine to use X-Ray when enabled tracingConfig.

(Should I rename commit type 'feat' to 'fix'?)

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -429,6 +429,13 @@ module.exports = {
         });
       }
 
+      if (stateMachineObj.tracingConfig) {
+        iamPermissions.push({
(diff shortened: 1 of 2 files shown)


horike37:
@hassaku63 
Thank you for fixing the IAM roles settings 👍
The code looks good from my perspective.

>(Should I rename commit type 'feat' to 'fix'?)

Would be better to rename the 'fix' type since that will automatically introduce the minor version up.
Can you update the commit message? Afterward. We can merge and release it.

theburningmonk (approved):
LGTM

hassaku63:
@horike37 Thanks for your review. I changed to "fix".

horike37:
@hassaku63 
Thank you for updating 👍

PR #380, opened 2020-10-14
fix: typo in executionsSucceeded

vbhavsar:
Per [AWS docs](https://docs.aws.amazon.com/step-functions/latest/dg/procedure-cw-metrics.html#cloudwatch-step-functions-execution-metrics), the correct name is ExecutionsSucceeded.


vbhavsar:
@theburningmonk Thanks again for the quick review and merge. Is it possible to get a (patch) version bump?

PR #401, opened 2021-01-26
Update README.md

owain68:
It is quite confusing the way the state machines have been named i.e. what is syntatical and what is just a name.  Additionally, the eventbridge should be split out as a separate section since the syntax is confusing since eventbridge is just a synonym of cloudwatch - it appears AWS are moving more to the EventBridge as a service name.


horike37 (approved):
Thanks!

PR #426, opened 2021-05-07
chore: bump asl-validator to fix cve-2021-23358

j0k3r:
CVE-2021-23358 affects the `underscore` library which is used by `jsonpath` in `asl-validator`.

`asl-validator` 1.8.* use `jsonpath` 1.0.* which use `underscore` 1.7.0, affected by the CVE.
`asl-validator` 1.9.* use `jsonpath` 1.1.* which use `underscore` 1.12.1, which is the earliest fixed version.

Also replace all `http://npm.paesslergmbh.de:4873` by `https://registry.npmjs.org` in `package-lock.json` because the url looked dead to me. I also don't understand why it was here...

Previous `package-lock.json` was generated by NPM 6, here I used NPM 7. That's why the whole file changed.


horike37 (approved):
@j0k3r 
Thank you for the report and fix 👍

PR #443, opened 2021-07-02
feat: support importing DynamoDB tables exported from external stacks

### LINKED ISSUE #442: Feature: Support importing table names using Fn::ImportValue when using the direct DynamoDB actions
# This is a Feature Proposal (and PR)

## Description
I want to be able to update data directly from a State Machine using [AWS' DynamoDB Step Function integration](https://docs.aws.amazon.com/step-functions/latest/dg/connect-ddb.html) and import the table name being updated from an external CloudFormation stack since that is where the table is defined, manged and exported, using [`Fn::ImportValue`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html).

e.g.

```yaml
name: example
definition:
  StartAt: Update My Externally Managed Table
  States:
    Update My Externally Managed Table:
      Type: Task
      Resource: "arn:aws:states:::dynamodb:updateItem"
      Parameters:
        TableName:
          Fn::ImportValue: MyExternalStack:Table:Name # <-- This
        ...
```

## Pull Request
I've already implemented this feature and manually tested it to verify it works correctly. I also added a unit test to verify the behaviour and prevent regression.

**PR:** #443 

## Testing

I did verify both in the IAM Web Console that the generated role was correct, as well as running my State Machines and seeing that the `UpdateItem` now succeeds whereas before it failed with insufficient permissions.

The generated IAM permission is as follows:

```json
{
    "Action": [
        "dynamodb:UpdateItem"
    ],
    "Resource": [
        "arn:aws:dynamodb:us-west-2:1234567890:table/imported-table-name"
    ],
    "Effect": "Allow"
}
```

After that I created a unit test to prevent regression.

danrivett:
Fixes #442

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ Specify your state machine definition using Amazon States Language in a `definit
 
 Alternatively, you can also provide the raw ARN, or SQS queue URL, or DynamoDB table name as a string. If you need to construct the ARN by hand, then we recommend to use the [serverless-pseudo-parameters](https://www.npmjs.com/package/serverless-pseudo-parameters) plugin together to make your life easier.
 
+In addition, if you want to reference a DynamoDB table managed by an external CloudFormation Stack, as long as that table name is exported as an output from that stack, it can be referenced by importing it using `Fn::ImportValue`. See the `ddbtablestepfunc` Step Function definition below for an example.
+
 ```yml
 functions:
   hello:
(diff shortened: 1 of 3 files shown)


horike37:
@danrivett 
This addition looks good. Thanks!
Can you add the example to README as well?

danrivett:
Thanks @horike37. I updated the `README.md` with an example, but let me know if you want me to make any changes. Thanks.

horike37:
@danrivett 
Thank you for the update.
Perfect 👍

PR #457, opened 2021-09-15
Regognize API Gateway properties at `provider.apiGateway`

medikoo:
In the context of Serverless Framework, API Gateway specific properties as `apiKeys`, `resourcePolicy`, `usagePlan` were moved from `provider` context to `provider.apiGateway` context.

There's no plan to support old notation with a new major release of the Framework.

This PR ensures that the plugin will remain working without issues in v3 of the Framework

/cc @horike37


horike37 (approved):
@medikoo 
Thank you for this fix! LGTM

Issue #377, opened 2020-09-19
Custom Authorizer Error


db46:
# This is a Bug Report

## Description

For bug reports:
I have added a custom authorizer as per the plugins instructions, however it repeatedly fails. I have tried many different combinations, including the exact arn for the authorizer lambda.

Using an IAM as an authorizer works - but this was only to validate that an authorizer could be added.

Here is an extract of my serverless.yml
```
custom:
  pythonRequirements:
    dockerizePip: non-linux
  authorizer: arn:aws:lambda:#{AWS::Region}:#{AWS::AccountId}:function:auth-service-${self:provider.stage}-auth

functions:
  stepTest:
    handler: stepTest.lambda_handler
    package: {}

stepFunctions:
  stateMachines:
    hellostepfunc1:
      events:
        - http:
            path: tenants
            method: post
            authorizer: ${self:custom.authorizer}
            cors: true
            request:
              template: lambda_proxy
      name: OnBoardingStateMachine
      definition:
        Comment: "Step function to manage the onboarding service"
        StartAt: HelloWorld1
        States:
          HelloWorld1:
            Type: Task
            Resource: arn:aws:lambda:#{AWS::Region}:#{AWS::AccountId}:function:${self:service}-${self:provider.stage}-stepTest
            InputPath: $
            Parameters:
              input.$: $
              step: HelloWorld1
            End: true
resources:
  Outputs:
    MyStateMachine:
      Description: The ARN of the example state machine
      Value:
        Ref: OnBoardingStateMachine
```

I get the following output:
```
  Type Error ---------------------------------------------
 
  TypeError: Cannot read property 'Fn::GetAtt' of undefined
      at getFunctionLogicalId 
```

Similar or dependent issues:
* #38 

## Additional Data
* ***Serverless Framework Core Version you're using***: 1.82.0
* ***The Plugin Version you're using***: 2.26
* ***Operating System***: MacOS

vladgolubev:
Facing the same issue as the OP. @db46 have you found a solution to this?

ereber:
I am also having this problem @db46 .  Is there a solution or a workaround?

PR #459, opened 2021-09-29
fix: arn authorizers

Danwakeem:
> This fixes issue #377 

## Description
Fixed an issue when using an arn for your authorizer rather than a function reference. I was getting the same issue when I was referencing a CloudFormation output variable and this PR fixes that case as well 😎


Danwakeem:
Side note: This is a fantastic plugin 🎸 so If there is anything else I can do to improve this PR please let me know 😎

horike37:
@Danwakeem 
Thank you for this fix! And great to hear that!
looks good from my perspective.

PR #461, opened 2021-10-07
start to working with us gov

calvinmetcalf:
my attempt to deploy a step function using this to a us gov region was failing due to some hard coded arn prefixes, this doesn't get all of them but it's a start that allows my app to deploy


horike37:
This feature has been supported on v3.2.1. Closing this issue.

PR #472, opened 2021-11-25
Mark plugin as Serverless Framework v3 ready

medikoo:
Plugin is confirmed to work with v3 release without issues, therefore it'll be good whitelist v3 of the Framework in peer dependencies section


horike37:
Thanks @medikoo !

medikoo:
Thank you @horike37 !

PR #481, opened 2021-11-25
docs: fix Inline Link in README.md

bayoudhi:
Hi, the current **Specify Input or Inputpath** inline link is not working.


bayoudhi:
I test it and it works in my forked repo:
https://github.com/bayoudhi/serverless-step-functions/blob/patch-1/README.md#specify-input-or-inputpath-or-inputtransformer

horike37:
Good catch! Thanks!

PR #485, opened 2021-12-15
feat: use pseudo region in role template

### LINKED ISSUE #407: Bug - Hardcoded Region breaks single package deployment strategies
# This is a Bug Report

## Description

For bug reports:
* What went wrong?

When packaging a stack once and deploying the artifact to multiple regions the AssumePolicyDocument contains the wrong region value.  Looks like [iam-role-statemachine-execution-template.txt](https://github.com/serverless-operations/serverless-step-functions/blob/master/lib/iam-role-statemachine-execution-template.txt#L10) is hardcoded to use the build time region not the runtime region. We publish one artifact at build time and then use our internal build tool to propagate that through our testing and production environments. Having a single region referenced there instead of a psuedo CloudFormation  parameter breaks that

* What did you expect should have happened?

Expectation: "states.[region].amazonaws.com"
Actual: "states.${AWS::Region}.amazonaws.com"

* What was the config you used?
Happens regardless of config.

## Additional Data
An easy fix to this should be to update [iam-role-statemachine-execution-template.txt](https://github.com/serverless-operations/serverless-step-functions/blob/master/lib/iam-role-statemachine-execution-template.txt#L10) to use 
"states.${AWS::Region}.amazonaws.com" instead of hard coding the region

Current workaround: 
```
Properties:
     AssumeRolePolicyDocument:
        Statement:
          - Effect: Allow
             Principal:
             Service:
                Fn::Sub:
                  - states.${AWS::Region}.amazonaws.com
                  - {}
              Action: sts:AssumeRole
```

laardee:
This PR changes the role template to use the pseudo region parameter.

fixes #407

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -547,7 +547,6 @@ module.exports = {
       );
 
       const iamRoleJson = iamRoleStateMachineExecutionTemplate
-        .replace('[region]', this.options.region)
         .replace('[PolicyName]', this.getStateMachinePolicyName())
         .replace('[Statements]', JSON.stringify(iamStatements));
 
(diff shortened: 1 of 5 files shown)


horike37:
Thanks @laardee !
LGTM.

laardee:
Nice! Thanks @horike37

PR #507, opened 2022-05-17
feat: add support for having a task be based on an object in S3

jswift:
Creates the IAM Role for tasks which are to fetch an object and return it's content, for example:

```yaml
FetchExecutionConfiguration:
    Type: Task
    Resource: arn:aws:states:::aws-sdk:s3:getObject
    Next: DoSomeThingWithConfiguration
    Parameters:
    Bucket: "my-execution-configs"
    Key: "config.json"
```

Currently a task defined this way will result in the following error.

> Cannot generate IAM policy statement for Task state

With this change it correctly deploys and runs, with only the minimum permission granted (i.e. `s3:GetObject` on the specified file).

See:
[Supported AWS SDK service integrations](https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html#supported-services-awssdk-list)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,35 @@
+name: Publish to NPM
+
+env:
+  package: serverless-step-functions
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build-and-publish:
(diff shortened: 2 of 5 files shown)


zirkelc:
Just stumbled upon the same issue, would be good to have this PR merged and have support for other operations besides `getObject` as well.

lopburny:
Hi @jswift, Thank you for this PR👍

Before merging it, would you remove some changes below that are not supposed to be related to this PR? :)

- .github/dependabot.yml
- .github/workflows/release.yml
- `package.json`

jswift:
Sorry about that, I've remade the pull request from a new branch that only contains this fix :)

pull request #516

PR #516, opened 2022-07-26
feat: add support for having a task be based on an object in S3

jswift:
Creates the IAM Role for tasks which are to fetch an object and return it's content, for example:

```yaml
FetchExecutionConfiguration:
    Type: Task
    Resource: arn:aws:states:::aws-sdk:s3:getObject
    Next: DoSomeThingWithConfiguration
    Parameters:
    Bucket: "my-execution-configs"
    Key: "config.json"
```

Currently a task defined this way will result in the following error.

> Cannot generate IAM policy statement for Task state

With this change it correctly deploys and runs, with only the minimum permission granted (i.e. `s3:GetObject` on the specified file).

See:
[Supported AWS SDK service integrations](https://docs.aws.amazon.com/step-functions/latest/dg/supported-services-awssdk.html#supported-services-awssdk-list)

Note: This is the fixed version of pull request #507

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -372,6 +372,15 @@ function getEventBridgePermissions(state) {
   ];
 }
 
+function getS3GetObjectPermissions(state) {
+  return [{
+    action: 's3:GetObject',
+    resource: [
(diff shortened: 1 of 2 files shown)


lopburny:
Thank you @jswift All looks good to me!

PR #525, opened 2022-08-26
fix: upgrading libs with vulnerabilities

rappanui:
SNYK found some vulnerabilities on nested dependencies in this project. This commit fix this


lopburny:
@rappanui Thank you for this PR. LGTM!

PR #528, opened 2022-09-21
feat: add support for Redshift Data

kmfukuda:
Adds support for Redshift Data.

It makes it possible to create state machines that directly use Redshift Data to run statements on clusters with temporary credentials and then fetch the results.


lopburny:
Hi @kmfukuda Thank you for this PR. All looks good to me!

PR #529, opened 2022-10-22
fix: Make it work with serverless-import-config-plugin

warrickhill:
We use serverless-import-config-plugin, but this plugin will only use config in the main serverless.yml entry file.

This PR fixes that and sets the defaults in a simpler way


horike37:
Thank you for this PR! We'll take a look.

lopburny:
Thank you @warrickhill Looks good to me!

PR #534, opened 2022-11-18
feature: update asl-validator to v3

nicksiscoe:
The old `asl-validator` doesn't support `aws-us-gov` (or, for that matter, anything other than just `aws`) as a partition when specifying the `Resource:` property in step functions

That makes it impossible to use this plugin with AWS GovCloud :(

Fortunately, upgrading the dependency to its latest version solves the issue!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@hapi/joi": "^15.0.2",
         "@serverless/utils": "^6.7.0",
-        "asl-validator": "^1.9.8",
+        "asl-validator": "^3.0.8",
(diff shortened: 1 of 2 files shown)


lopburny:
@nicksiscoe Thanks for this PR. Hopefully this will make it available to you folks!

lopburny:
@nicksiscoe Hi, the semantic-release bot says it was released in 3.10.2, but I had to adjust the latest version for some reason, so now it is 3.11.1. please find the update from the latest 3.11.1, which should be included. Thanks!

nicksiscoe:
Thank you for the fast turnaround @lopburny ❤️

PR #535, opened 2022-11-24
fix(permissionsBoundary): apply permissionsBoundary

DubKF:
This PR allows using a Permissions Boundary defined in the provider level e.g. 

``` yaml
provider:
  name: aws
  runtime: python3.9
  # the permisions boundary goes here 
  rolePermissionsBoundary: arn:aws:iam::[account]:policy/permission_boundary  
  # ...
```

Previously, the boundary was only applied to the created lambda execution role, but was missing on the created roles for functions of the state machine as described in [issue 395](https://github.com/serverless-operations/serverless-step-functions/issues/395) some time ago. 

We run into the same issue yesterday well using the newest version of serverless and this plugin. The submitted fix worked for us.

### DIFF
--- a/lib/deploy/events/schedule/compileScheduledEvents.js
+++ b/lib/deploy/events/schedule/compileScheduledEvents.js
@@ -5,6 +5,8 @@ const BbPromise = require('bluebird');
 
 module.exports = {
   compileScheduledEvents() {
+    const service = this.serverless.service;
+    const permissionsBoundary = service.provider.rolePermissionsBoundary;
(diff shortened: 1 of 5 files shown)


DubKF:
@lopburny Could you perhaps have a look and provide some guidance or workflow approval as I'm a 1st contributor ;)

lopburny:
Hi @DubKF Thanks for this PR. Looks good to me.

nabsource:
please add this option to the docs/readme

PR #536, opened 2022-12-06
feat: support the latest SFN ItemProcessor (replaces Iterator)

bahrmichael:
When building a new step function I noticed that what was previously the `Iterator` has been changed to `ItemProcessor`. Probably changed with reinvent and their release of Distributed Maps. Here's an example of a state machine that I built with their visual builder and then adjusted to reference my functions: https://github.com/bahrmichael/trade-game-backend/pull/50/commits/9a75d378d7e67673bffc448f5507793c2eb9408e#diff-ea2dbe8c0992878214f1020b0c7ee79668dc63b5c8325a5ed67892d0080a9b27

When setting `validate: true` I noticed that the plugin expects an `Iterator` within a `Map` task, which fails with the new `ItemProcessor`: https://github.com/bahrmichael/trade-game-backend/actions/runs/3633513042/jobs/6130610217

This PR shows how far I've gotten, but there seems to be another error where I'm not sure if merging this PR already would cause problems for other folks: https://github.com/bahrmichael/trade-game-backend/actions/runs/3633603701/jobs/6130803277

Let me know what you think! I'm not confident that the test is sufficient, please give me a pointer on what it should also test.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -19,7 +19,7 @@ function getTaskStates(states) {
         return getTaskStates(parallelStates);
       }
       case 'Map': {
-        const mapStates = state.Iterator.States;
+        const mapStates = state.ItemProcessor ? state.ItemProcessor.States : state.Iterator.States;
         return getTaskStates(mapStates);
       }
       default: {
--- a/lib/deploy/stepFunctions/compileStateMachines.test.js
+++ b/lib/deploy/stepFunctions/compileStateMachines.test.js
@@ -1647,4 +1647,85 @@ describe('#compileStateMachines', () => {
     // Definition is invalid and validate=true, should throw
     expect(() => serverlessStepFunctions.compileStateMachines()).to.throw(Error);
   });
+
+  it('should compile with the old Iterator', () => {
+    serverless.service.stepFunctions = {
+      stateMachines: {
+        myStateMachine1: {
+          id: 'Test',
+          name: 'test',
+          definition: {
+            StartAt: 'One',
+            States: {
+              One: {
+                Type: 'Map',
+                Iterator: {
+                  StartAt: 'Two',
+                  States: {
+                    Two: {
+                      Type: 'Wait',
+                      Seconds: 10,
+                      End: true,
+                    },
+                  },
+                },
+                End: true,
+              },
+            },
+          },
+        },
+      },
+    };
+
+    serverlessStepFunctions.compileStateMachines();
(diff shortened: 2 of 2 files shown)


lopburny:
Hi @bahrmichael Thank you for this PR. Let me take a look just in case and I'll update you soon🙌

JeffBeltran:
I ran into this same issue. Thanks @bahrmichael for getting this started.

After a bit more digging, it appears the [Iterator state is now deprecated](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#map-state-inline-additional-fields)

> The `ItemProcessor` field replaces the now deprecated [Iterator](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#iterator) field. Although you can continue to include `Map` states that use the `Iterator` field, we highly recommend that you replace this field with `ItemProcessor`

[then more info as well](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#map-state-inline-deprecated-fields)

> Although you can continue to include `Map` states that use the following fields, we highly recommend that you replace `Iterator` with [ItemProcessor](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#itemprocessor) and `Parameters` with [ItemSelector](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#itemselector).

When i generate a new step function i just copy paste the generated code from the aws console. Since they use the newer `ItemProcessor` by default, I just went and changed it to use the deprecated `Iterator` key and this fixed the deployment issue on my side. 

```
      "Type": "Map",
      "ItemProcessor": {
        "ProcessorConfig": {
          "Mode": "INLINE"
        },
        "StartAt":...
```
to 

```
      "Type": "Map",
      "Iterator": {
        "ProcessorConfig": {
          "Mode": "INLINE"
        },
        "StartAt": ....
```

That being said, it would be nice to get this merged as eventually this workaround will not work

j0k3r (approved):
I guess you can squash your commits

bahrmichael:
@j0k3r Done! Do you think the tests are sufficient? I'm not sure how I can test this plugin locally.

j0k3r:
@bahrmichael at least it's seems ok to me! :)

bahrmichael:
Awesome! I'll be the first to use it once it's merged :)

bahrmichael:
@lopburny Could I get your eyes on this again?

lopburny:
Thank you @bahrmichael @j0k3r , All look good to me👍

PR #539, opened 2022-12-27
Adding support for Map in Distributed Mode

brycenrogers:
AWS released a new feature for Map called Distributed Mode, which allows for distributed processing in a Map. The IAM role compilation needs to be tweaked to handle this new format, otherwise it deploys fine.

Note: this does not currently work with `validate: true` as [asl-validator](https://github.com/ChristopheBougere/asl-validator) does not yet support this behavior.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -19,8 +19,13 @@ function getTaskStates(states) {
         return getTaskStates(parallelStates);
       }
       case 'Map': {
-        const mapStates = state.Iterator.States;
-        return getTaskStates(mapStates);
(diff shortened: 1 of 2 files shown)


harazdovskiy:
Hi, @brycenrogers, I'm also lacking DISTRIBUTED Map feature. To understand you correctly, you have closed this pull because of [asl-validator](https://github.com/ChristopheBougere/asl-validator)?

brycenrogers:
Hey @harazdovskiy I closed this because it was implemented right before I created this PR in https://github.com/serverless-operations/serverless-step-functions/pull/536. You should be able to define the Distributed Map without issue in the next release. asl-validator does still need to update their logic to handle the new ItemProcessor/Distributed Map feature though.

PR #545, opened 2023-01-12
feat: generate IAM policy for S3 when ItemReader is used

akshaydk:
**Issue**: When `ItemReader` is used to download the files from S3, IAM policy is not created to access S3.

**Solution**: Adding an IAM policy with `Resource: *` when the bucket and key are sent dynamically.

 `Resource: *`: [ItemReader AWS Documenation](https://docs.amazonaws.cn/en_us/step-functions/latest/dg/input-output-itemreader.html#itemreader-iam-policiesac)

**Notes**: I cross referenced the policy by  creating step function from AWS UI and the IAM policy generated had `Resource: *` when `bucket` and `key` were passed dynamically.

Please let me know if I need to make any further changes.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -20,7 +20,11 @@ function getTaskStates(states) {
       }
       case 'Map': {
         const mapStates = state.ItemProcessor ? state.ItemProcessor.States : state.Iterator.States;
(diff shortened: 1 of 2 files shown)


lopburny:
@akshaydk Thanks for this PR. LGTM!

bahrmichael:
Thank you @akshaydk !

PR #546, opened 2023-01-12
feat: add IAM role to distributed_map to start the state machine

akshaydk:
**Issue**: When executing a state machine with a [Map state in Distributed mode](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-distributed.html), the execution fails at the Map state because execution role does not have sufficient permissions to run a Distributed Map.

**Solution**: I tired to use the current pass that we do to collect the tasks and generate additional policy which gives access to trigger the `stepFunction`. I couldn't find a better way to access the name of the `stepFunction` so I had to pass it as a parameter.

This should also solve this [issue](https://github.com/serverless-operations/serverless-step-functions/issues/543)

**Notes:**

- With the changes in the PR, I've deployed my stepFunction to AWS and it seems to work alright in distributed mode.
- Please let me know if I can make it better

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -8,7 +8,7 @@ const { getArnPartition } = require('../../utils/arn');
 
 const logger = require('../../utils/logger');
 
-function getTaskStates(states) {
+function getTaskStates(states, stateMachineName) {
   return _.flatMap(states, (state) => {
     switch (state.Type) {
       case 'Task': {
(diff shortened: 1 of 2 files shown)


lopburny:
Hi @akshaydk Thank you for this PR and looks good to me!
There seems to be a conflict related to #545 , could you look over it and resolve it? I'll merge this PR and release it right after it is resolved!

akshaydk:
Hi @lopburny You can review the code now.

lopburny:
@akshaydk LGTM!

akshaydk:
@lopburny Thanks for the review, will you be releasing a new version today?

lopburny:
*Deleting the branch for this PR does not affect this repository

bahrmichael:
Thank you @akshaydk !

PR #550, opened 2023-01-16
fix: template error while using distributed_map

akshaydk:
Issue: 
```The CloudFormation template is invalid: Template error: One or more Fn::Sub intrinsic functions don't specify expected arguments. Specify a string as first argument, and an optional second argument to specify a mapping of values to replace in the string```

This PR fixes the above issue by adding the second argument.


lopburny:
@akshaydk Looks good. I'll release this fix again!

lopburny:
@akshaydk Released! Could you check if it works now?

PR #552, opened 2023-01-30
feat: support custom state machine names when using a distributed mapA

### LINKED ISSUE #543: Missing IAM Role permissions for Distributed Map states
# This is a Bug Report

## Description
* **What went wrong?**  
 
(Preface: In this issue I used an unreleased version of this plugin which incorporates the changes from #536)  

When executing a state machine with a [Map state in Distributed mode](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-distributed.html), the execution fails at the Map state because execution role does not have sufficient permissions to run a Distributed Map.

This is due to the required additional permissions, where the execution role for the state machine containing the Distributed Map must be able to `StartExecution` on itself: https://docs.aws.amazon.com/step-functions/latest/dg/iam-policies-eg-dist-map.html

* **What did you expect should have happened?**
 
The Distributed Map state should have run without permissions issues.  

* **What was the config you used?**

Not really applicable here, just a standard distributed map state:
```
MyMapState:
  Type: Map
  ItemProcessor:
    ProcessorConfig:
      Mode: DISTRIBUTED
      ExecutionType: STANDARD
    StartAt: FooState
    States:
      FooState:
        ....
```

* **What stacktrace or error message from your provider did you see?**

This is the error that appears on the AWS Dashboard during the execution of the state machine:
```
User: arn:aws:sts::12345:assumed-role/MyStateMachineRole/abcdef is not authorized to perform: states:StartExecution on resource: arn:aws:states:us-east-1:12345:stateMachine:MyStateMachineName because no identity-based policy allows the states:StartExecution action (Service: Sfn, Status Code: 400, Request ID: some-uuid)
``` 

## Additional Data
* ***Serverless Framework Core Version you're using***: 3.25.1
* ***The Plugin Version you're using***: Unreleased; my package.json points to this commit: https://github.com/serverless-operations/serverless-step-functions/commit/42bd42350c33c418762bff6f75cb8b2d6e2d3729
* ***Operating System***: N/A
* ***Stack Trace***: N/A
* ***Provider Error messages***:  N/A (see description above)

I've been able to brute-force a fix by adding the following rule to the set of IAM Permissions:
https://github.com/timorthi/serverless-step-functions/commit/fd1b36a534ec9fa64b9df6ef940b9a889458d423#diff-5efbd2d24990cd41d11040072f12d980a6ae28e66faef40a046e8d30d69ff528R580-R590

There is probably a better/cleaner way to add the permissions based on the detection of a Distributed Map state, however it appears that `getTaskStates` ignores "container" states like Maps and Parallels so there's no easy way to detect a Distributed Map without changing `getTaskStates`.

toddhainsworth:
Fixes #543 by looking for a 'name' field on the state machine and falling back to using it's ID if no name is provided.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -597,14 +597,15 @@ module.exports = {
     logger.config(this.serverless, this.v3Api);
     const service = this.serverless.service;
     const permissionsBoundary = service.provider.rolePermissionsBoundary;
-    this.getAllStateMachines().forEach((stateMachineName) => {
-      const stateMachineObj = this.getStateMachine(stateMachineName);
+    this.getAllStateMachines().forEach((stateMachineId) => {
+      const stateMachineObj = this.getStateMachine(stateMachineId);
+      const stateMachineName = stateMachineObj.name || stateMachineId;
       if (stateMachineObj.role) {
         return;
       }
 
       if (!stateMachineObj.definition) {
-        throw new Error(`Missing "definition" for state machine ${stateMachineName}`);
+        throw new Error(`Missing "definition" for state machine ${stateMachineId}`);
       }
 
       const taskStates = getTaskStates(stateMachineObj.definition.States, stateMachineName);
@@ -646,7 +647,7 @@ module.exports = {
       }
 
       const stateMachineLogicalId = this.getStateMachineLogicalId(
-        stateMachineName,
+        stateMachineId,
         stateMachineObj,
       );
       const iamRoleStateMachineLogicalId = `${stateMachineLogicalId}Role`;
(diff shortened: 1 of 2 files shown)


toddhainsworth:
Hey @lopburny is there any chance this could be reviewed and included in the next release? This bug is currently blocking some current and upcoming work for us.

akshaydk (approved):
LGTM

toddhainsworth:
Hey @akshaydk do you happen to know when the next release is planned? Keen to upgrade to unblock a bunch of work we're doing :pray:

akshaydk:
@toddhainsworth I'm sorry. I just a contributor to this project. 

@lopburny Can you take a look at this?

lopburny:
@toddhainsworth @akshaydk Thank you so much to you folks! All looks good to me👍

PR #562, opened 2023-04-14
Issue #562

jappurohit041:
# Individual Step Function Deploy

This document outlines the process for adding the individual step function deploy feature to a plugin.

## Background

Step functions are a powerful tool for building serverless workflows. They allow you to orchestrate multiple AWS services into a cohesive workflow. However, deploying step functions can be challenging. AWS provides a web-based interface for building and testing step functions, but deploying them requires the use of the AWS CLI or SDK. This can be complex and error-prone, especially when you have multiple step functions to deploy.

## Problem

The current plugin for deploying step functions does not allow for individual step function deployment. This means that all step functions in a stack must be deployed at once, even if only one has changed. This can be time-consuming and can increase the risk of errors.

## Solution

The solution is to add an individual step function deploy feature to the plugin. This feature will allow developers to deploy a single step function without having to deploy the entire stack. This will save time and reduce the risk of errors.

## Implementation

To implement the individual step function deploy feature, the following steps should be taken:

1. Add a new command to the plugin that allows for the deployment of a single step function. This command should take the name of the step function as a parameter.
2. Modify the existing deploy command to allow for the deployment of a step functions at once. This commands take a name of step function as a parameter.
3. Sample argument: 

```java
serverless deploy stepf --name *Step_Machine_Name*
```

## Conclusion

By adding the individual step function deploy feature to the plugin, developers will have greater control over the deployment of their step functions. This will save time and reduce the risk of errors.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,33 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch Program",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${workspaceFolder}/lib/index.js"
+        },
+        {
+          "type": "node",
+          "name": "Mocha Current File",
+          "request": "launch",
+          "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+          "args": [
+            "--timeout",
+            "999999",
+            "--colors",
+            "${file}"
+          ],
+          "console": "integratedTerminal",
+          "internalConsoleOptions": "neverOpen",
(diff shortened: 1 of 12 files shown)


lopburny:
Hi @jappurohit041 thank you for your work for this PR. I had taken a look and I am wondering if it might cause issue with the existing stack since it seems it directly calls AWS SDK for updating the individual state machine. please let me know what you think.

jappurohit041:
We checked that did not faced any such issues in cloud formation updation...

lopburny:
Hi @jappurohit041 Sorry for the silence! We've been discussing this topic internally for a while and concluded that it would be good to set up an issue and get people's feedback on updating process using SDK. I know that your team have been working on this for some time, so how about using it as a local plugin first, and then finalize the spec on the issue? I'm willing to speak up as well...!

lopburny:
If you re-post this PR in the issue, I will comment that I've reviewed the demo and summarize our team's concerns so that we can discuss further there 🙌

KillDozerX2:
Hi @lopburny, any updates on this? Saw a few PRs being merged over the past week.

PR #567, opened 2023-05-02
Chore/566 add apigateway property to provider in documentation example

tvhees:
This PR does not include functional changes but corrects documentation which would otherwise lead to a deployment error if followed.


lopburny:
@tvhees LGTM!

PR #576, opened 2023-08-01
feat: expand support for Redshift Data

kmfukuda:
Expands support for Redshift Data.

It adds support for running statements on Redshift Serverless workgroups as well as using Secrets Manager secrets or IAM identities for authentication.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -245,25 +245,82 @@ function getBatchDynamoDBPermissions(action, state) {
 }
 
 function getRedshiftDataPermissions(action, state) {
+  const permissions = [];
+
   if (['redshift-data:ExecuteStatement', 'redshift-data:BatchExecuteStatement'].includes(action)) {
-    const clusterName = _.has(state, 'Parameters.ClusterIdentifier') ? state.Parameters.ClusterIdentifier : '*';
-    const dbName = _.has(state, 'Parameters.Database') ? state.Parameters.Database : '*';
-    const dbUser = _.has(state, 'Parameters.DbUser') ? state.Parameters.DbUser : '*';
-    return [{
+    const dbName = _.has(state, ['Parameters', 'Database']) ? state.Parameters.Database : '*';
+
+    let workgroupArn;
+    let clusterName;
+    if (_.has(state, ['Parameters', 'WorkgroupName'])) {
+      if (state.Parameters.WorkgroupName.startsWith('arn:')) {
+        workgroupArn = state.Parameters.WorkgroupName;
+      } else {
+        workgroupArn = { 'Fn::Sub': 'arn:${AWS::Partition}:redshift-serverless:${AWS::Region}:${AWS::AccountId}:workgroup/*' };
+      }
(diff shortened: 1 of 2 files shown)


kmfukuda:
Hi @lopburny Is there any chance that this could be included in the next release?

lib/deploy/stepFunctions/compileIamRole.js
```diff
@@ -245,25 +245,82 @@ function getBatchDynamoDBPermissions(action, state) {
...
+        permissions.push({
+          action: 'redshift:GetClusterCredentialsWithIAM',
+          resource: { 'Fn::Sub': `arn:\${AWS::Partition}:redshift:\${AWS::Region}:\${AWS::AccountId}:dbname:${clusterName}/${dbName}` },
+        });
+      }
+    }
+  } else {
+    permissions.push({
+      action,
+      resource: '*',
```
horike37, line 319: How do you reach this section under what conditions? Does this mean this section add * permission to the generated IAM by this plugin?

kmfukuda: @horike37 Thank you for the review.

This statement is executed when the normalized Resource field value of a Task state is one of the following ARNs.

- `arn:aws:states:::aws-sdk:redshiftdata:listStatements`
- `arn:aws:states:::aws-sdk:redshiftdata:describeStatement`
- `arn:aws:states:::aws-sdk:redshiftdata:getStatementResult`
- `arn:aws:states:::aws-sdk:redshiftdata:cancelStatement`

The corresponding actions do not support resource-level permissions, therefore requiring the wildcard `*` to be specified as the resource in the policy ([documentation](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonredshiftdataapi.html#amazonredshiftdataapi-actions-as-permissions)). The resulting statement would for example look like this.

```
{
  "Effect": "Allow",
  "Action": [
    "redshift-data:ListStatements"
  ],
  "Resource": "*"
}
```

The `getRedshiftDataPermissions` function, which includes the code in question, is called by the `getIamPermissions` function from within lines 607-618, which are not changed in this pull request.

horike37: Thank you for the explanation! I got it.

horike37:
@kmfukuda 
Thank you for this PR. Looks good to me. 

Have you done actual tests for this feature? I mean like if this feature doesn't introduce any side effects to existing features, or if this keeps backward compatibility. Because I couldn't have enough time to check it right now.

If so, I will merge this PR and release it

kmfukuda:
I deployed and executed some state machines for testing using the code before and after the change, and they all behaved as intended.

To be specific, I deployed and executed a state machine for each of the only actions `redshift-data:ExecuteStatement` and `redshift-data:BatchExecuteStatement` that can result in different permissions because of the change and each of parameter sets that collectively enable covering lines 251-315, and the state machines showed that the code after the change supports all the cases that the code before the change supports.

I have also confirmed that the other actions that are handled by `getRedshiftDataPermissions` within lines 317-320 support the newly supported cases.

horike37:
Thank you for the explanation. Perfect!

PR #580, opened 2023-08-23
Create Schedules using Eventbridge Scheduler

### LINKED ISSUE #571: EventBridge Scheduler Integration for Stepfunction
# Feature Proposal

For feature proposals:
-> Ability to create AWS::Scheduler::Schedule triggers as an alternative to AWS::Event::Rule, which may be preferable if hitting account-wide limits on a number of AWS::Event::Rule triggers per bus (Lambdas are only able to be triggered from the default bus).
-> Also, specify a timezone to run it on local time.

KillDozerX2:
Fixes #571

### DIFF
--- a/lib/deploy/events/schedule/compileScheduledEvents.js
+++ b/lib/deploy/events/schedule/compileScheduledEvents.js
@@ -121,8 +121,9 @@ module.exports = {
 
             const stateMachineLogicalId = this
               .getStateMachineLogicalId(stateMachineName, stateMachineObj);
-            const scheduleLogicalId = this
-              .getScheduleLogicalId(stateMachineName, scheduleNumberInFunction);
+            const scheduleLogicalId = method !== METHOD_SCHEDULER ? this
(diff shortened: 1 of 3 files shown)


jappurohit041:
Hii @horike37 @lopburny,

Could we please have look at this PR, as it has been opened since very long and it is kind of important to make it in sync with Serverless https://github.com/serverless/serverless/pull/11935 ?

Thanks

horike37:
@jappurohit041 
Thank you for this PR. Looks good to me, seeing the added code and test that checks if the required parameters are given from serverless.yml.

Have you done actual tests for this feature? I mean like if this feature doesn't introduce any side effects to existing features, or if this keeps backward compatibility. Because I couldn't have enough time to check it.

If so, I will merge this PR and release it after you fix some linting errors.

jappurohit041:
Hii @horike37,

I have fixed lint issue, and due to the distinct logical ID, two test failed, so fixed the test case.

Also, I have done the actual tests for this feature, after the latest commit, and it is not breaking.

Working as expected😄

PR #581, opened 2023-08-23
Individual Step Function Deploy

jappurohit041:
# Individual Step Function Deploy

This document outlines the process for adding the individual step function deploy feature to a plugin.

## Background

Step functions are a powerful tool for building serverless workflows. They allow you to orchestrate multiple AWS services into a cohesive workflow. However, deploying step functions can be challenging. AWS provides a web-based interface for building and testing step functions, but deploying them requires the use of the AWS CLI or SDK. This can be complex and error-prone, especially when you have multiple step functions to deploy.

## Problem

The current plugin for deploying step functions does not allow for individual step function deployment. This means that all step functions in a stack must be deployed at once, even if only one has changed. This can be time-consuming and can increase the risk of errors.

## Solution

The solution is to add an individual step function deploy feature to the plugin. This feature will allow developers to deploy a single step function without having to deploy the entire stack. This will save time and reduce the risk of errors.

## Implementation

To implement the individual step function deploy feature, the following steps should be taken:

1. Add a new command to the plugin that allows for the deployment of a single step function. This command should take the name of the step function as a parameter.
2. Modify the existing deploy command to allow for the deployment of a step functions at once. This commands take a name of step function as a parameter.
3. Sample argument: 

```java
serverless deploy stepf --name *Step_Machine_Name*
```

## Conclusion

By adding the individual step function deploy feature to the plugin, developers will have greater control over the deployment of their step functions. This will save time and reduce the risk of errors.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,33 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch Program",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${workspaceFolder}/lib/index.js"
+        },
+        {
+          "type": "node",
+          "name": "Mocha Current File",
+          "request": "launch",
+          "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+          "args": [
+            "--timeout",
+            "999999",
+            "--colors",
+            "${file}"
+          ],
+          "console": "integratedTerminal",
+          "internalConsoleOptions": "neverOpen",
+          "skipFiles": ["<node_internals>/**"],
+          "envFile": "${workspaceFolder}/.env"
+        }
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "conventionalCommits.scopes": [
+        "Individual Deploy"
+    ]
+}
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Serverless Framework v2.32.0 or later is required.
  - [Tags](#tags)
  - [Commands](#commands)
      - [deploy](#deploy)
+     - [deploy individual step function](#deploy_individual_step_function)
      - [invoke](#invoke)
  - [IAM Role](#iam-role)
  - [Tips](#tips)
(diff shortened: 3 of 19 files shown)


jappurohit041:
[Fix](https://github.com/serverless-operations/serverless-step-functions/issues/570)

jappurohit041:
Hii @horike37 @lopburny,

Could we please have look at this PR, as it has been opened since very long?

Thanks

horike37:
@jappurohit041 
Thank you for this PR! I understand what problem should be solved by this PR.
However, in my opinion, this change would be risky in terms of the consistency of Cloudformation. we've got a possibility of breaking a Cloudformation stack in some cases. One of the key concepts of the Serverless Framework is keeping the consistency of deployment. The reason why only one Cloudformation stack is deployed from one serverless.yml fim. 

Moreover, this project is not actually active. so I don't want to introduce huge changes related to Cloudformation's behavior.

How about releasing this feature as an isolated plugin so that you can control this feature and keep maintaining for a long time?

jappurohit041:
Hii @horike37,

Thank you for the feedback!
I do understand your point but I would like say that, modifying definition string will not impact or break cloud-formation stack. Here we are using UpdateStateMachine [https://docs.aws.amazon.com/step-functions/latest/apireference/API_UpdateStateMachine.html] from SDK to do individual deploy. Secondary, we have validation in place to prevent from inconsistency the cloudformation. 

Also, to do a individual deploy there is separate command which will be acting as enhance feature without breaking existing flow. Yes, here you can see many changes, but those are due test-service which is created for testing the plugin. If you want we can remove it.

Lastly, having isolated plugin, will cause too much confusion for the community, and unnecessary maintenance  of two plugin. If you are willing and comfortable to handover, than my team and me would be more than happy to maintain this plugin as it is very critical plugin.

If you want, we can discuss offline as well :).. Do let me know your thoughts

Thanks

jappurohit041:
> > Lastly, having isolated plugin, will cause too much confusion for the community, and unnecessary maintenance of two plugin. If you are willing and comfortable to handover, than my team and me would be more than happy to maintain this plugin as it is very critical plugin.
> 
> That seems really interesting. I've just decided to make a fork to use it myself because the low maintainance. @jappurohit041 you can drop your thoughts here #587

@ebisbe That really great 😄... I hope this feature would have proved to be very useful for you....

horike37:
@jappurohit041 
Thank you for your response. I understand what you stated. No offense but it seems that no one supports for this feature other than you so far based on https://github.com/serverless-operations/serverless-step-functions/issues/570.

To be honest, I am a bit skeptical of the necessity of this feature whether many users actually want it, or not. Moreover, you mentioned that this feature should be included in this plugin so that this wouldn't cause too much confusion for the community. I don't think so. We've got the official plugin repository and this is maintained by the Serverless,inc team on https://github.com/serverless/plugins. You can easily find the plugins that you need and it's normal that you can create any plugins if you want and put them in the repository.

MitanshGor:
Absolutely, integrating the individual step function deploy feature is paramount. It will bolsters efficiency by eliminating the requirement to deploy the whole stack for minor adjustments. This will surely diminishes error risks and  greater control to developers, I feel it will enhance developer experience and overall efficiency. These changes will undoubtedly be highly beneficial and significantly improve the developer experience and efficiency.

anjalimodi10:
We already know the advantage of deploying a single lambda function rather than deploying the entire cloudformation stack. We used to miss this functionality for deploying a single step function. I hope the current PR will resolve that and enhances our developers life in a better way.

ebisbe:
Now that I'm using more step functions that would be something handy to have.

Issue #370, opened 2020-09-09
No typescript support


LL782:
Serverless Version: 1.78.1
Plugin Version: 2.22.1

Let's have a Typescript types for serverless-step-functions

**Why**

1. ts-check fails when using `serverless-step-functions` because `stepFunctions` is not expected by [Serverless defintion](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/serverless/plugins/aws/provider/awsProvider.d.ts)
1. Types would make work quicker and easier in VS Code and other IDEs that support Typescript
1. It would also give human readers a good overview of `stepFunctions`

**What**

- Something like this [definition for Serverless](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/serverless/plugins/aws/provider/awsProvider.d.ts) from the **[@types/serverless](https://www.npmjs.com/package/@types/serverless) npm**

**Notes**

I have started doing this in my own project. See an example for my specific use case below

I'm raising this issue to see where/how/if we can develop a complete definition collectively.

```ts
// serverless.ts
import { Serverless } from "serverless/aws";
import { contentfulEnvironmentVariables } from "./src/config/contenful";

type Definition = {
  Comment?: string;
  StartAt: string;
  States: {
    [state: string]: {
      Catch?: Catcher[];
      Type: "Map" | "Task" | "Choice" | "Pass";
      End?: boolean;
      Next?: string;
      ItemsPath?: string;
      ResultPath?: string;
      Resource?: string | { "Fn::GetAtt": string[] };
      Iterator?: Definition;
    };
  };
};

type Catcher = {
  ErrorEquals: ErrorName[];
  Next: string;
  ResultPath?: string;
};

type ErrorName =
  | "States.ALL"
  | "States.DataLimitExceeded"
  | "States.Runtime"
  | "States.Timeout"
  | "States.TaskFailed"
  | "States.Permissions"
  | string;

interface ServerlessWithStepFunctions extends Serverless {
  stepFunctions: {
    stateMachines: {
      [stateMachine: string]: {
        name: string;
        definition: Definition;
      };
    };
    activities?: string[];
    validate?: boolean;
  };
}

// example config (in a Typescript Serverless project this can be used in place of serverless.yml)

const serverlessConfiguration: ServerlessWithStepFunctions = {
  service: {
    name: "xxx",
  },
  frameworkVersion: ">=1.72.0",
  custom: {
    webpack: {
[...]

PR #585, opened 2023-09-07
Add TypeScript types

zirkelc:
This PR adds TypeScript types as discussed in #370 

First, you need to use the official TypeScript [@serverless/typescript](https://github.com/serverless/typescript#readme) package. for your `serverless.ts` file.

TypeScript merges the types from this package with the base types from @serverless/typescript with either of these two options:

- Add `/// <reference types="serverless-step-functions" />` at the top of your `serverless.ts`
```ts
/// <reference types="serverless-step-functions" />
import type { AWS as Serverless } from '@serverless/typescript';
const serverlessConfiguration: Serverless = {
  // ...
};
```

- Add `"types": ["serverless-step-functions"]` to your `tsconfig.json`
```json5
{
  "compilerOptions": {
    // ...
    "types": [
      "serverless-step-functions"
    ]
  },
  "include": [
    "src/**/*.ts",
    "serverless.ts",
  ],
}
```

I can't guarantee that the types are fully complete and correct, but I think there are a good starting point.

If you rather prefer to have the types provided as a separate `@types/serverless-step-functions` package, I'd be willing to submit a PR to the [DefinetlyTypes](https://github.com/DefinitelyTyped/DefinitelyTyped) package.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -10,13 +10,15 @@ declare module '@serverless/typescript' {
 }
 
 type StateMachines = {
-  [stateMachine: string]: {
-    type?: 'STANDARD' | 'EXPRESS'; 
-    id?: string;
-    name?: string;
-    definition: Definition;
-    tracingConfig?: TracingConfig;
-  };
+  [stateMachine: string]: StateMachine
+};
+
+export type StateMachine = {
+  type?: 'STANDARD' | 'EXPRESS';
+  id?: string;
+  name?: string;
+  definition: Definition;
+  tracingConfig?: TracingConfig;
 };
 
 type Selector = string | Record<string, unknown>;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -141,6 +141,7 @@ type ChoiceRuleOr = {
 
 type ChoiceRuleSimple = ChoiceRuleComparison & {
   Next: string;
+  Comment?: string:
 };
 
 type ChoiceRule = ChoiceRuleSimple | ChoiceRuleNot | ChoiceRuleAnd | ChoiceRuleOr;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -158,7 +158,7 @@ interface Map extends StateBase {
   Type: 'Map';
   ItemsPath: string;
   Iterator: Definition;
-  ItemSelector: {
+  ItemSelector?: {
     [key: string]: string;
   }
   MaxConcurrency?: number;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -5,6 +5,7 @@ declare module '@serverless/typescript' {
     stepFunctions?: {
       stateMachines: StateMachines;
       validate?: boolean;
+      noOutput?: boolean;
     };
   }
 }
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -43,6 +43,9 @@ type Definition = {
   Comment?: string;
   StartAt: string;
   States: States;
+  ProcessorConfig?: {
+    Mode: "INLINE" | "DISTRIBUTED";
+  };
 };
 
 type States = {
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -130,6 +130,7 @@ type ChoiceRuleNot = {
 type ChoiceRuleAnd = {
   And: ChoiceRuleComparison[];
   Next: string;
+  Comment?: string;
 };
 
 type ChoiceRuleOr = {
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -145,6 +145,9 @@ interface Map extends StateBase {
   Type: 'Map';
   ItemsPath: string;
   Iterator: Definition;
+  ItemSelector: {
+    [key: string]: string;
+  }
   MaxConcurrency?: number;
   Parameters?: {
     [key: string]: string;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -136,6 +136,7 @@ type ChoiceRuleAnd = {
 type ChoiceRuleOr = {
   Or: ChoiceRuleComparison[];
   Next: string;
+  Comment?: string;
 };
 
 type ChoiceRuleSimple = ChoiceRuleComparison & {
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -11,6 +11,7 @@ declare module '@serverless/typescript' {
 
 type StateMachines = {
   [stateMachine: string]: {
+    type?: 'STANDARD' | 'EXPRESS'; 
     id?: string;
     name?: string;
     definition: Definition;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -124,6 +124,7 @@ type ChoiceRuleComparison = {
 type ChoiceRuleNot = {
   Not: ChoiceRuleComparison;
   Next: string;
+  Comment?: string;
 };
 
 type ChoiceRuleAnd = {
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -11,7 +11,7 @@ declare module '@serverless/typescript' {
 
 type StateMachines = {
   [stateMachine: string]: {
-    id: string;
+    id?: string;
     name: string;
     definition: Definition;
     tracingConfig?: TracingConfig;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -156,7 +156,7 @@ interface Fail extends StateBase {
 
 interface Map extends StateBase {
   Type: 'Map';
-  ItemsPath: string;
+  ItemsPath?: string;
   Iterator: Definition;
   ItemSelector?: {
     [key: string]: string;
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,241 @@
+import { AWS } from '@serverless/typescript';
+
+declare module '@serverless/typescript' {
+  interface AWS {
+    stepFunctions?: {
+      stateMachines: StateMachines;
+      validate?: boolean;
+    };
+  }
+}
+
+type StateMachines = {
+  [stateMachine: string]: {
+    id: string;
+    name: string;
+    definition: Definition;
+    tracingConfig?: TracingConfig;
+  };
+};
+
+type Selector = string | Record<string, unknown>;
+
+type Input = {
+  'AWS_STEP_FUNCTIONS_STARTED_BY_EXECUTION_ID.$'?: string;
+} & Record<string, any>;
+
+type ResultSelector = Record<string, any>;
+
+type Parameter = {
+  'Payload.$'?: string;
+  Input?: Input;
+  FunctionName?: string;
+  StateMachineArn?: string;
+} & {
+  [selector: string]: Selector;
+};
+
+type TracingConfig = {
+  enabled: boolean;
+};
+
+type Definition = {
+  Comment?: string;
+  StartAt: string;
+  States: States;
+};
+
+type States = {
+  [state: string]: Choice | Fail | Map | Task | Parallel | Pass | Wait | Succeed;
+};
+
+type StateBase = {
+  Catch?: Catcher[];
+  Retry?: Retrier[];
+  End?: boolean;
+  InputPath?: string;
+  Next?: string;
+  OutputPath?: string;
+  ResultPath?: string | null;
+  ResultSelector?: { [key: string]: string | { [key: string]: string } };
+  Type: string;
+  Comment?: string;
+};
+
+type ChoiceRuleComparison = {
+  Variable: string;
+  BooleanEquals?: boolean;
+  BooleanEqualsPath?: string;
+  IsBoolean?: boolean;
+  IsNull?: boolean;
+  IsNumeric?: boolean;
+  IsPresent?: boolean;
+  IsString?: boolean;
+  IsTimestamp?: boolean;
+  NumericEquals?: number;
+  NumericEqualsPath?: string;
+  NumericGreaterThan?: number;
+  NumericGreaterThanPath?: string;
+  NumericGreaterThanEquals?: number;
+  NumericGreaterThanEqualsPath?: string;
+  NumericLessThan?: number;
+  NumericLessThanPath?: string;
+  NumericLessThanEquals?: number;
+  NumericLessThanEqualsPath?: string;
+  StringEquals?: string;
+  StringEqualsPath?: string;
+  StringGreaterThan?: string;
+  StringGreaterThanPath?: string;
+  StringGreaterThanEquals?: string;
+  StringGreaterThanEqualsPath?: string;
+  StringLessThan?: string;
+  StringLessThanPath?: string;
+  StringLessThanEquals?: string;
+  StringLessThanEqualsPath?: string;
+  StringMatches?: string;
+  TimestampEquals?: string;
+  TimestampEqualsPath?: string;
+  TimestampGreaterThan?: string;
+  TimestampGreaterThanPath?: string;
+  TimestampGreaterThanEquals?: string;
+  TimestampGreaterThanEqualsPath?: string;
+  TimestampLessThan?: string;
+  TimestampLessThanPath?: string;
+  TimestampLessThanEquals?: string;
+  TimestampLessThanEqualsPath?: string;
+};
+
+type ChoiceRuleNot = {
+  Not: ChoiceRuleComparison;
+  Next: string;
+};
+
+type ChoiceRuleAnd = {
+  And: ChoiceRuleComparison[];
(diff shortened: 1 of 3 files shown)


ebisbe:
Aren't `id` and `name` not mandatory?

lib/index.d.ts
```diff
@@ -0,0 +1,241 @@
...
+interface Fail extends StateBase {
+  Type: 'Fail';
+  Cause?: string;
+  Error?: string;
+}
+
+interface Map extends StateBase {
+  Type: 'Map';
+  ItemsPath: string;
+  Iterator: Definition;
```
ebisbe, line 144: Missing: 
```
ItemSelector?: {
    [key: string]: string | Array<unknown> | { [key: string]: string };
  }
```

lib/index.d.ts
```diff
@@ -0,0 +1,241 @@
...
+    stepFunctions?: {
+      stateMachines: StateMachines;
+      validate?: boolean;
+    };
+  }
+}
+
+type StateMachines = {
+  [stateMachine: string]: {
+    id: string;
```
ebisbe, line 14:
```suggestion
    id?: string;
```

lib/index.d.ts
```diff
@@ -0,0 +1,241 @@
...
+    name: string;
```
ebisbe, line 15:
```suggestion
    name?: string;
```

lib/index.d.ts
```diff
@@ -0,0 +1,241 @@
...
+};
+
+type TracingConfig = {
+  enabled: boolean;
+};
+
+type Definition = {
+  Comment?: string;
+  StartAt: string;
+  States: States;
```
ebisbe, line 45:
```suggestion
  States: States;
  ProcessorConfig?: {
    Mode: "INLINE" | "DISTRIBUTED";
  };
```

lib/index.d.ts
```diff
@@ -0,0 +1,241 @@
...
+    };
+  }
+}
+
+type StateMachines = {
+  [stateMachine: string]: {
+    id: string;
+    name: string;
+    definition: Definition;
+    tracingConfig?: TracingConfig;
```
ebisbe, line 17:
```suggestion
    tracingConfig?: TracingConfig;
    events?: any;
    dependsOn?: any;
```

Not really sure what would be their types.

lib/index.d.ts
ebisbe, line 144:
```suggestion
  Iterator: Definition;
  ItemSelector: {
    [key: string]: string;
  }
```

ebisbe (commented):
That's what I've found missing with my current step function I'm building.

ebisbe:
Also when I try to export the state machine to a single file it fails with:

```
Cannot load "serverless.ts": Initialization error: TSError: ⨯ Unable to compile TypeScript:
serverless.ts:650:7 - error TS2322: Type '{ events: { schedule: { rate: string; enabled: boolean; }; }[]; dependsOn: string[]; definition: { StartAt: string; States: { "Add empty LastEvaluatedKey": { Type: string; Next: string; Result: { LastEvaluatedKey: string; }; ResultPath: string; }; ... 4 more ...; "Add new LastEvaluatedKey": { ...; }; }; }; }' is not assignable to type '{ id?: string; name?: string; events?: any; dependsOn?: any; definition: Definition; tracingConfig?: TracingConfig; }'.
  The types of 'definition.States' are incompatible between these types.
    Type '{ "Add empty LastEvaluatedKey": { Type: string; Next: string; Result: { LastEvaluatedKey: string; }; ResultPath: string; }; Scan: { Type: string; Next: string; Parameters: { TableName: { Ref: string; }; FilterExpression: string; ExpressionAttributeNames: { ...; }; ExpressionAttributeValues: { ...; }; "ExclusiveStart...' is not assignable to type 'States'.
      Property '"Add empty LastEvaluatedKey"' is incompatible with index signature.
        Type '{ Type: string; Next: string; Result: { LastEvaluatedKey: string; }; ResultPath: string; }' is not assignable to type 'Choice | Fail | Map | Task | Parallel | Pass | Wait | Succeed'.
          Type '{ Type: string; Next: string; Result: { LastEvaluatedKey: string; }; ResultPath: string; }' is not assignable to type 'Pass'.
            Types of property 'Type' are incompatible.
              Type 'string' is not assignable to type '"Pass"'.

650       addCredits,
          ~~~~~~~~~~

  node_modules/serverless-step-functions/lib/index.d.ts:13:3
     13   [stateMachine: string]: {
          ~~~~~~~~~~~~~~~~~~~~~~~~~
     14     id?: string;
        ~~~~~~~~~~~~~~~~
    ... 
     19     tracingConfig?: TracingConfig;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     20   };
        ~~~~
    The expected type comes from this index signature.
    ```

File: lib/index.d.ts (hunk 2 above)
Comment (ebisbe) [inline lib/index.d.ts:14]:
```suggestion
    id: string;
    type?: 'STANDARD' | 'EXPRESS'; 
```

lib/index.d.ts
```diff
@@ -0,0 +1,241 @@
...
+  interface AWS {
+    stepFunctions?: {
+      stateMachines: StateMachines;
+      validate?: boolean;
+    };
+  }
+}
+
+type StateMachines = {
+  [stateMachine: string]: {
```
ebisbe, line 13:
```suggestion
  [stateMachine: string]: {
    type?: 'STANDARD' | 'EXPRESS'; 
```

lib/index.d.ts
zirkelc: Looking at the source code:
https://github.com/serverless-operations/serverless-step-functions/blob/28fa07b254bd2f6fef934fe940911dd36fa77615/lib/deploy/stepFunctions/compileStateMachines.schema.js#L37-L40
https://github.com/serverless-operations/serverless-step-functions/blob/28fa07b254bd2f6fef934fe940911dd36fa77615/lib/deploy/stepFunctions/compileStateMachines.schema.js#L63

```suggestion
    tracingConfig?: TracingConfig;
    events?: any[];
    dependsOn?: string | string[];
```

zirkelc:
> Also when I try to export the state machine to a single file it fails with:
> 

What do you mean with exporting it to a single file? The types use module augmentation to merge the types into the official AWS type:

https://github.com/serverless-operations/serverless-step-functions/blob/c464dcdc0e0255cd93170486c7fd907b9893c020/lib/index.d.ts#L3-L10

If you want to use the types as stand-alone, we should probably export the `stepFunctions` type separately.

Is that what you mean?

ebisbe:
> > Also when I try to export the state machine to a single file it fails with:
> 
> What do you mean with exporting it to a single file? The types use module augmentation to merge the types into the official AWS type:
> 
> https://github.com/serverless-operations/serverless-step-functions/blob/c464dcdc0e0255cd93170486c7fd907b9893c020/lib/index.d.ts#L3-L10
> 
> If you want to use the types as stand-alone, we should probably export the `stepFunctions` type separately.
> 
> Is that what you mean?

Nope. What I wanted to have in a different file is the stateMachine implementation. Not the types. So I can have a file for each StateMachine and split things so I don't have a huge serverless.ts file. 

I've also found that we are missing `type?: 'STANDARD' | 'EXPRESS'; ` and `loggingConfig?: any;` ( https://github.com/serverless-operations/serverless-step-functions/issues/281 )

zirkelc:
> > > Also when I try to export the state machine to a single file it fails with:
> > 
> > 
> > What do you mean with exporting it to a single file? The types use module augmentation to merge the types into the official AWS type:
> > https://github.com/serverless-operations/serverless-step-functions/blob/c464dcdc0e0255cd93170486c7fd907b9893c020/lib/index.d.ts#L3-L10
> > 
> > If you want to use the types as stand-alone, we should probably export the `stepFunctions` type separately.
> > Is that what you mean?
> 
> Nope. What I wanted to have in a different file is the stateMachine implementation. Not the types. So I can have a file for each StateMachine and split things so I don't have a huge serverless.ts file.
> 

Yes, I think the issue wouldn't appear if you could type your stateMachine like this:

```ts
export const stateMachine1: StateMachine = { 
  // ...
};
```

I assume the issue is related to [type widening](https://mariusschulz.com/blog/literal-type-widening-in-typescript) by TypeScript. So the last error message as an example:
> Types of property 'Type' are incompatible.
>  Type 'string' is not assignable to type '"Pass"'.

The state machine expects `type` to be a string literal of `Pass` or `Task` etc. But when you define your state machine outside of the serverless type, TypeScript widens the type from literal `Pass` to `string`. That causes the type error I guess.

ebisbe:
> I assume the issue is related to [type widening](https://mariusschulz.com/blog/literal-type-widening-in-typescript) by TypeScript. So the last error message as an example:
> 
> > Types of property 'Type' are incompatible.
> > Type 'string' is not assignable to type '"Pass"'.
> 
> The state machine expects `type` to be a string literal of `Pass` or `Task` etc. But when you define your state machine outside of the serverless type, TypeScript widens the type from literal `Pass` to `string`. That causes the type error I guess.

Ok, I didn't know the term or what was really happening but that's exactly the issue. Then what we could do is export the single type `StateMachine`

lib/index.d.ts
```diff
@@ -0,0 +1,248 @@
...
+    type?: 'STANDARD' | 'EXPRESS'; 
+    id?: string;
+    name?: string;
+    definition: Definition;
+    tracingConfig?: TracingConfig;
+  };
+};
```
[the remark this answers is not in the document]
zirkelc: Yes, that should work. Can you give it a try?

lib/index.d.ts
```diff
@@ -0,0 +1,250 @@
...
+type StateMachines = {
+  [stateMachine: string]: StateMachine
+};
+
+export type StateMachine = {
+  type?: 'STANDARD' | 'EXPRESS';
+  id?: string;
+  name?: string;
+  definition: Definition;
+  tracingConfig?: TracingConfig;
```
ebisbe, line 21:
```suggestion
  tracingConfig?: TracingConfig;
  loggingConfig?: LoggingConfig;
  events?: any[];
  dependsOn?: string | string[];
};

type LoggingConfig = {
  level: "ERROR" | "ALL" | "FATAL" | "OFF";
  includeExecutionData: boolean;
  destinations: Resource | Resource[]
```
We forgot that one in one subcomment.

lib/index.d.ts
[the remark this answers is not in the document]
ebisbe: That's working.

ebisbe:
Everything is working from my side. Not sure if we are missing anything.

lib/index.d.ts
```diff
@@ -0,0 +1,259 @@
...
+
+interface Fail extends StateBase {
+  Type: 'Fail';
+  Cause?: string;
+  Error?: string;
+}
+
+interface Map extends StateBase {
+  Type: 'Map';
+  ItemsPath: string;
```
ebisbe, line 158:
```suggestion
  ItemsPath?: string;
```

lib/index.d.ts
```diff
@@ -0,0 +1,259 @@
...
+  Iterator: Definition;
+  ItemSelector: {
```
ebisbe, line 160:
```suggestion
  ItemSelector?: {
```

lib/index.d.ts
```diff
@@ -0,0 +1,259 @@
+import { AWS } from '@serverless/typescript';
+
+declare module '@serverless/typescript' {
+  interface AWS {
+    stepFunctions?: {
+      stateMachines: StateMachines;
+      validate?: boolean;
```
ebisbe, line 7:
```suggestion
      validate?: boolean;
      noOutput?: boolean
```

lib/index.d.ts
```diff
@@ -0,0 +1,260 @@
...
+  Next: string;
+};
+
+type ChoiceRuleOr = {
+  Or: ChoiceRuleComparison[];
+  Next: string;
+};
+
+type ChoiceRuleSimple = ChoiceRuleComparison & {
+  Next: string;
```
ebisbe, line 140:
```suggestion
  Next: string;
  Comment?: string:
```

lib/index.d.ts
```diff
@@ -0,0 +1,260 @@
...
+};
+
+type ChoiceRuleAnd = {
+  And: ChoiceRuleComparison[];
+  Next: string;
+};
+
+type ChoiceRuleOr = {
+  Or: ChoiceRuleComparison[];
+  Next: string;
```
ebisbe, line 136:
```suggestion
  Next: string;
  Comment?: string;
```

lib/index.d.ts
```diff
@@ -0,0 +1,260 @@
...
+};
+
+type ChoiceRuleNot = {
+  Not: ChoiceRuleComparison;
+  Next: string;
+};
+
+type ChoiceRuleAnd = {
+  And: ChoiceRuleComparison[];
+  Next: string;
```
ebisbe, line 131:
```suggestion
  Next: string;
  Comment?: string;
```

lib/index.d.ts
```diff
@@ -0,0 +1,260 @@
...
+  TimestampGreaterThanEqualsPath?: string;
+  TimestampLessThan?: string;
+  TimestampLessThanPath?: string;
+  TimestampLessThanEquals?: string;
+  TimestampLessThanEqualsPath?: string;
+};
+
+type ChoiceRuleNot = {
+  Not: ChoiceRuleComparison;
+  Next: string;
```
ebisbe, line 126:
```suggestion
  Next: string;
  Comment?: string;
```

zirkelc:
@horike37 could you please take a look if you have the chance. If you prefer not to include TS types in this package, we can move this PR to the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.

lib/index.d.ts
```diff
@@ -0,0 +1,264 @@
...
+  | TaskParametersForSNS
+  | { [key: string]: string | { [key: string]: string } };
+}
+
+interface Pass extends StateBase {
+  Type: 'Pass';
+  Parameters?: {
+    [key: string]: string | Array<unknown> | { [key: string]: string };
+  };
+  Result?: {
```
ebisbe, line 219:
```suggestion
  Result?: string | {
```
```suggestion
  Result?: string | [] | {
```
I'm using `Result: "${env:something}"` to add an array. Not sure what's the best approach here. The string value resolves into an array. The tip from the Ui console says: `Must be valid JSON.` Array and object are the only ones valid.

horike37:
@zirkelc 
Thank you for this PR! Adding the type system should be really important. But right now we can't regularly maintain this repo, so it would be great if you can move this PR to the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.

lib/index.d.ts
zirkelc: Should we try something like this: https://github.com/microsoft/TypeScript/issues/1897#issuecomment-331765301

```ts
export type JSONPrimitive = string | number | boolean | null;
export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
export type JSONObject = { [member: string]: JSONValue };
export interface JSONArray extends Array<JSONValue> {}
```

```suggestion
  Result?: JSONValue;
```

zirkelc:
@horike37 thanks for you reply! Alright, I'll submit a PR to DefinitelyTyped in the next few days.

zirkelc:
@ebisbe I created a new PR in the DefinetlyTyped repo: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66693 

I didn't try it yet, but I think to get the types merged from this package, we have to use module augmentation in our `serverless.ts` file directly:

```
import type { AWS } from '@serverless/typescript';
import type { StepFunctions } from 'serverless-step-functions';

declare module '@serverless/typescript' {
   interface AWS {
     stepFunctions?: StepFunctions;
  }
}
```

ebisbe:
> we have to use module augmentation

I don't know what's that it. I will try to help you on the other PR.

zirkelc:
It means add types from package `serverless-step-functions` into the base types of `@serverless/typescript`. However, I figured out that type packages can install other type packages. So we get it to work without add this part at the top:

```
import type { AWS } from '@serverless/typescript';
import type { StepFunctions } from 'serverless-step-functions';

declare module '@serverless/typescript' {
   interface AWS {
     stepFunctions?: StepFunctions;
  }
}
```

We have to wait for the approval on the DT repo.

ebisbe:
Ok. What I don't like on this PR is that we have been doing it manually... It would be perfect if we could extract the full state machine definition and convert it to types so it can be automated. Here it's for CF resources https://github.com/awboost/cfntypes ( only sadly )

ebisbe:
@zirkelc this package has a validation option. Let me see what is using under the hood.

PR #594, opened 2023-10-29
[Issue-593] Change selectpattern from int to string

louis0802:
Fixing for the issue https://github.com/serverless-operations/serverless-step-functions/issues/593

### DIFF
--- a/lib/deploy/events/apiGateway/methods.js
+++ b/lib/deploy/events/apiGateway/methods.js
@@ -196,13 +196,13 @@ module.exports = {
       IntegrationResponses: [
         {
           StatusCode: 200,
-          SelectionPattern: 200,
+          SelectionPattern: '200',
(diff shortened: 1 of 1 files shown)


louis0802:
Hii @horike37 @lopburny,

Could we please have a look at this PR?

Thanks

horike37:
@louis0802 
Thank you for this PR. No one seems to face this issue. I'm concerned that this would introduce any side effects.
Could you check if it breaks the backward compatibility?

louis0802:
> @louis0802 Thank you for this PR. No one seems to face this issue. I'm concerned that this would introduce any side effects. Could you check if it breaks the backward compatibility?

@horike37 
May I have your help to check if it breaks the backward compatibility?
Is there any script or command I could run to verify it?

Thanks

PR #597, opened 2023-11-17
docs: update README with EventBridge Scheduler

cm-uehara:
The functionality of EventBridge Scheduler has been added to solve the issue #571, and the method to utilize this functionality has been documented.


horike37:
@cm-uehara 
Good catch! Thank you for the update.

PR #598, opened 2023-11-23
feat: add support for step func sdk sync execution

vcarity:
This PR adds support to generate IAM statements for ```StartSyncExecution``` when using a nested express workflow.

While working with a express stepFunction I needed to generate permissions for a nested express workflow like this: 

```Resource: arn:aws:states:::aws-sdk:sfn:startSyncExecution```. 

After verifying the plugin I noticed it only works for:

```
arn:aws:states:::states:startExecution
arn:aws:states:::states:startExecution.sync
arn:aws:states:::states:startExecution.sync:2
arn:aws:states:::states:startExecution.waitForTaskToken
```
The .sync, .sync:2 and waitForTaskToken are not supported by express workflows.

With this PR we should be able to add the necessary action ```states:StartSyncExecution```.
All I did was to replicate the ```states:startExecution``` for ```aws-sdk:sfn:startSyncExecution```

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -440,6 +440,39 @@ function getStepFunctionsPermissions(state) {
   }];
 }
 
+function getStepFunctionsSDKPermissions(state) {
+  let stateMachineArn = state.Mode === 'DISTRIBUTED' ? {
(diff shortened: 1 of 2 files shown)


joonatassouza:
@horike37 Please, if u could review this, i'm also in need and that will solve our problem

PR #600, opened 2024-01-10
Align policy naming with serverless framework convention?

kazdy:
Hi, 

I noticed that policies created by serverless-step-functions plugin does not conform to the serverless framework naming convention.

Example serverless framework generated policy name:
`abc-serverless-dev-lambda`
vs one created with serverless-step-functions plugin:
`dev-abc-serverless-statemachine`

I tend to use the service name to restrict on what resources my deployment IAM role can operate by checking for `${service-name}-*` and this behavior breaks this convention that serverless framework follows.

I don't know if this behavior is intentional or not, but I would like to propose this change and see what others think about it.

### DIFF
--- a/lib/naming.js
+++ b/lib/naming.js
@@ -35,8 +35,8 @@ module.exports = {
 
   getStateMachinePolicyName() {
     return [
-      this.provider.getStage(),
       this.provider.serverless.service.service,
+      this.provider.getStage(),
       'statemachine',
     ].join('-');
   },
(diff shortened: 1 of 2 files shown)


horike37:
@kazdy 

Thank you for proposing the fix. We didn't have the intention not to follow the rule of the Serverless Framework.
It would be better if it could follow the rule, but there is one concern. Is it possible to break the backward compatibility with this PR? If there is someone doing something using my plugin's naming rule, the project would be possibly broken when they upgrade this plugin.

kazdy:
Hi, yes I think this can break stuff if someone adjusted their deployment role for how this plugin builds policy names.
Would it be ok to introduce a feature flag and keep things as is until the time comes to introduce breaking change, providing an alternative still?
Do you think it's worth doing so?

PR #605, opened 2024-03-04
feat: added support for IAM Permission of tasks that create EventBrid…

### LINKED ISSUE #604: Support EventBridge Scheduler Service integration
# This is a Feature Proposal

## Description

Support Generation of proper permissions for this integration
* A Task that creates an EventBridge Scheduler Schedule fails to create IAM Statement in IAM Role.
* The `arn:aws:states:::aws-sdk:scheduler:createSchedule` resource should be supported

Similar or dependent issues:
* #432

jormaechea:
Resolves #604


jormaechea:
Thanks @horike37!

PR #608, opened 2024-03-10
Ad support to s3:listObjectsV2

jcarle:
I encountered a permissions issue when trying to loop through files in a bucket within a stepfunctions map.
The following PR solves the problem.
Feel free to improve or correct it.

Thanks for your great work


horike37:
@jcarle 
Thank you for this PR. Would you add a test code to check if the added code works fine?

jcarle:
Hi @horike37
I already added a test code

Thanks!!

PR #613, opened 2024-06-10
fix: define proper permissions for index with Ref table

ebisbe:
When the "TableName" is a Ref the 'Arn' to use the index is not generated correctly:

```arn:aws:dynamodb:eu-west-1:679049160385:table/[object Object]/index/GsiPkState```

It should be:

```
{'Fn::Join': ['/', [{ 'Fn::GetAtt': ['hello', 'Arn'] }, 'index/GSI1']] }
```


ebisbe:
@horike37 Can you review this?

PR #615, opened 2024-06-14
chore: :arrow_up: upgrade serverless peer dependency

### LINKED ISSUE #614: Upgrade serverless peer dependency to v4
# This is a Feature Proposal

## Description

Since Serverless V4 became generally available since June 13th, I'd like to see the serverless peer dependency upgraded to accept v4.

* What is the use case that should be solved. The more detail you describe this in the easier it is to understand for us.  
  * Allow users of this lib to use the latest version of serverless.

threes-was-taken:
Closes #614 

## Changed
Upgraded serverless peer dependency to allow versions starting from 2.32.0 (equal to and/or greater than).

## Good to notice
1) When I ran `npm install`, I got the warning that this lock-file was made with an older version of npm. According to the CLI logs, some "supplemental" metadata was fetched from the registry.  
My system currently runs **node v20.14.0 (LTS)** and **npm v10.7.0**  

2) All tests succeeded after this change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,50 +1,86 @@
 {
   "name": "serverless-step-functions",
   "version": "3.11.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 3,
   "requires": true,
-  "dependencies": {
-    "2-thenable": {
-      "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


fernandohonig:
Hi, any progress here? I can't upgrade to sls4 until this is merged. Thanks

PR #631, opened 2024-11-03
fix(deps): bump asl-validator version to address jsonpath-plus vulnerability

minjie0501:
The asl-validator package relies on jsonpath-plus as a peer dependency, which contained a known vulnerability (details at https://nvd.nist.gov/vuln/detail/CVE-2024-21534).

This PR resolves the vulnerability by upgrading asl-validator to the latest version, which includes an updated, secure version of jsonpath-plus.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@serverless/utils": "^6.7.0",
-        "asl-validator": "^3.8.0",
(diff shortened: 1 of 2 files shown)


ynishimura:
Why hasn't this been released?

If it gets flagged during vulnerability checks and isn't being maintained, we'll need to consider whether to continue using it.

zirkelc:
I will close this PR since the asl-validator is already on ^3.11.0 in master.

PR #648, opened 2025-04-19
fix: added support for generating S3 permissions using Bucket references

### LINKED ISSUE #647: Incorrect IAM Permissions generated when S3 Bucket Reference used
## Summary
I'm trying to do an S3 `putObject` using and S3 SDK service integration that references an S3 bucket created in the same `serverless.yml` file.

I tried to reference the bucket's name using `Ref: bucketResource` something like as follows:

```yml
Write to an S3 Bucket defined in serverless.yml:
  Type: Task
  Resource: arn:aws:states:::aws-sdk:s3:putObject
  Parameters:
    Bucket:
      Ref: myBucketReference
    Key: Foo
    Body: Bar
  End: true
```

But it fails to generate the correct IAM policy definition. It generates something like:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": [
                "s3:PutObject"
            ],
            "Resource": [
                "arn:aws:s3:::[object Object]/*"
            ],
            "Effect": "Allow"
        }
    ]
}
```

(Notice the `[object Object]` instead of the resolved S3 bucket name).

Whereas it generates the correct IAM permissions when I invoke a Lambda function that references a Lambda function defined in the same `serverless.yml file.

```yml
Invoke Lambda defined in serverless.yml:
  Type: Task
  Resource:
    Fn::GetAtt: [myLambdaResource, Arn]
```

## Possible Root Cause
From my investigation it looks like the `getS3ObjectPermissions()` function in `compileIamRole.js` ([here](https://github.com/serverless-operations/serverless-step-functions/blob/67ddb591c31b5f63bfb86ae94fe9659b123ec9bd/lib/deploy/stepFunctions/compileIamRole.js#L635)) is too simplistic and just assumes the `bucket` field specified is a string literal rather than possibly being a reference.

As I mentioned above, references and other intrinsic functions such as `Fn::GetAtt` are supported for other resources, but it seems the IAM permissions generation for S3 resources is currently lacking this.

## Possible Solution
I modified the `getS3ObjectPermissions()` function locally to implement a possible solution, and it seems to work for me. I have submitted PR #648  for consideration.

I am happy to modify it as needed, or have it superseded by a better solution, but it would be great to fix this.

danrivett:
Resolves #647

Note: I'll add unit tests if the proposed solution in this PR is correct. This solution works for me, but I wasn't sure if there was a better way to fix this issue.

### DIFF
--- a/lib/deploy/stepFunctions/compileIamRole.js
+++ b/lib/deploy/stepFunctions/compileIamRole.js
@@ -634,6 +634,25 @@ function getEventBridgeSchedulerPermissions(state) {
   ];
 }
 
+// Because the S3 Bucket parameter can be either a literal name or a reference to an
+// existing S3 bucket we need to resolve
+function resolveS3BucketReference(bucket, resource) {
+  if (isIntrinsic(bucket)) {
+    return {
+      'Fn::Sub': [
+        resource,
+        { bucket },
+      ],
+    };
+  }
+
+  return resource.replaceAll('${bucket}', bucket);
+}
+
+function resolveS3BucketReferences(bucket, resources) {
+  return resources.map(resource => resolveS3BucketReference(bucket, resource));
+}
+
 function getS3ObjectPermissions(action, state) {
   const bucket = state.Parameters.Bucket || '*';
   const key = state.Parameters.Key || '*';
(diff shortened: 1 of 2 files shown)


lib/deploy/stepFunctions/compileIamRole.js
```diff
@@ -642,27 +660,27 @@ function getS3ObjectPermissions(action, state) {
     return [
       {
         action: 's3:Get*',
-        resource: [
-          `arn:aws:s3:::${bucket}`,
-          `arn:aws:s3:::${bucket}/*`,
-        ],
+        resource: resolveS3BucketReferences(bucket, [
+          `arn:aws:s3:::\${bucket}`,
```
danrivett, line 664: I decided to keep the strings as template literals and just escape the `${bucket}` so it wasn't interpolated here, but in the `resolveS3BucketReferences()` function.

This is because other affected strings below uses other variables such as `${prefix}` and `${key}` and so they needed to remain template literals to have those interpolated correctly.

So I decided for consistency I would keep them all as template literals and uniformly escape the `${bucket}`.

zirkelc:
It looks good. Could you add a test so we can see the compiled IAM template?

danrivett:
> It looks good. Could you add a test so we can see the compiled IAM template?

That's great. Sorry for the delay here. I'm currently under a project deadline squeeze for the next 3 weeks, but I'll get to it once that we've hit that release milestone unless someone else wants to add one ahead of that.

lib/deploy/stepFunctions/compileIamRole.js
danrivett: After review, I changed my mind and decided to use unescaped single quoted strings with a comment as I felt it was easier for the reader to understand why `bucket` wasn't being interpolated directly here.

zirkelc:
@danrivett thanks for adding the tests, they look good to me!

PR #747, opened 2026-04-02
fix(notifications): merge SNS/SQS policies for shared topics and queues (#275)

### LINKED ISSUE #275: Using notifications overwrites exisiting resouce permissions.
# This is a Bug Report

## Description

When using a pre-existing SNS topic for notifications, any existing policies on the topic are replaced with a new one.
If, like us, you have multiple alarms & notifications using the same queue (for fan-in) then it stop alarms from being able to publish to the topic.

Expected behaviour:
* The existing policy is be amended, rather than replaced.

Similar or dependent issues:
* none afaik (nothing in issues)

## Additional Data
* https://github.com/horike37/serverless-step-functions/blob/master/lib/deploy/stepFunctions/compileNotifications.js#L89 shows how the policy is generated
* https://github.com/horike37/serverless-step-functions/blob/master/lib/deploy/stepFunctions/compileNotifications.js#L264 being where it is built
(I'm sure you know that...)

VirtueMe:
## Summary

Fixes issue #275: when multiple state machines send notifications to the same SNS topic or SQS queue (fan-in), the plugin generates a separate `AWS::SNS::TopicPolicy` / `AWS::SQS::QueuePolicy` resource per state machine. `AWS::SNS::TopicPolicy` and `AWS::SQS::QueuePolicy` are **full replacements** in CloudFormation — when two resources target the same topic/queue, the second overwrites the first. The first machine's notification rules then silently stop working because their principal is no longer in the policy.

## Fix

After compiling all notification resources, same-target SNS topic policies and SQS queue policies are merged into a single CloudFormation resource with all statements combined. CloudFormation then applies the policy exactly once with all principals authorised.

The change is in `compileNotifications.js`: a `mergeResourcePolicies` step groups `AWS::SNS::TopicPolicy` resources by topic and `AWS::SQS::QueuePolicy` resources by queue, merging statements before the resources are written into the CloudFormation template.

## Verification

**Integration test:** The `notifications` fixture was updated to include a second state machine (`notificationMachine2`) sharing the same SNS topic and SQS queue as the first — the exact fan-in scenario from #275.

The compiled CloudFormation template was inspected after packaging with the fix applied:

```
SNS TopicPolicy resources: 1
  IntegrationDashnotificationsDashtestResourcePolicyc30a6: 2 statement(s)

SQS QueuePolicy resources: 1
  IntegrationDashnotificationsDashtestResourcePolicy6fbc7: 2 statement(s)
```

One merged policy per target, with one statement per state machine. Without the fix, two separate resources would be generated for each target, causing CloudFormation to overwrite the first machine's policy with the second's. LocalStack deploy is green.

**Unit tests:** Three new tests added covering:
- Two state machines sharing the same SNS topic → 1 policy, 10 merged statements
- Two state machines sharing the same SQS queue → 1 policy, 10 merged statements
- One state machine with the same topic on multiple statuses → 1 policy, 2 merged statements

## Test plan

- [ ] `npm test` — 533 passing
- [ ] `npx osls notifications:deploy --stage test` — green
- [ ] Packaged template confirms 1 merged `AWS::SNS::TopicPolicy` and 1 merged `AWS::SQS::QueuePolicy` with statements from both state machines

Closes #275

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/fixtures/notifications/serverless.yml
+++ b/fixtures/notifications/serverless.yml
@@ -29,3 +29,22 @@ stepFunctions:
         SUCCEEDED:
           - sqs:
               Fn::GetAtt: [NotificationQueue, Arn]
+
+    # Second state machine sharing the same SNS topic and SQS queue (fan-in).
+    # Reproduces issue #275: without the fix, this machine's TopicPolicy and
+    # QueuePolicy overwrite the first machine's, leaving it unable to publish.
+    notificationMachine2:
+      name: integration-notifications-2-${opt:stage, 'test'}
+      definition:
+        StartAt: PassThrough
+        States:
+          PassThrough:
+            Type: Pass
+            End: true
+      notifications:
+        FAILED:
+          - sns:
+              Ref: NotificationTopic
+        SUCCEEDED:
+          - sqs:
+              Fn::GetAtt: [NotificationQueue, Arn]
--- a/fixtures/notifications/verify.test.js
+++ b/fixtures/notifications/verify.test.js
@@ -1,0 +1,74 @@
+'use strict';
+
+const fs = require('node:fs');
+const path = require('node:path');
+const expect = require('chai').expect;
+
+const templatePath = path.join(__dirname, '.serverless', 'cloudformation-template-update-stack.json');
+
+describe('notifications fixture — CloudFormation template', () => {
+  let resources;
+
+  before(() => {
+    const template = JSON.parse(fs.readFileSync(templatePath, 'utf8'));
+    resources = template.Resources;
+  });
+
+  it('should produce exactly one AWS::SNS::TopicPolicy per unique topic', () => {
+    const snsPolicies = Object.values(resources).filter(
+      (r) => r.Type === 'AWS::SNS::TopicPolicy',
+    );
+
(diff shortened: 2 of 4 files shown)


VirtueMe:
Note: the manual template inspection used to verify this fix (#275) revealed a gap in the integration test suite — LocalStack deploys green for this class of bug but the incorrect template is only visible by inspecting the compiled output. Tracked in #748.

hfhchan-plb:
I think this is now causing `Invalid parameter: Every policy statement must have a unique ID` after an upgrade, any idea how to solve this issue?

VirtueMe:
> Invalid parameter: Every policy statement must have a unique ID

I will take a look into it!

VirtueMe:
@hfhchan-plb thanks for the report — you're right, and I've root-caused it. The merged policy ends up with duplicate statement `Sid`s because the `Sid` is derived only from status + target-type + (action, resource) and carries no state-machine identity, so two machines sharing a topic/queue emit identical statements that AWS then rejects with `Every policy statement must have a unique ID`.

Tracked in #801 with the fix (dedupe identical statements during the merge). Working on it now.

Issue #748, opened 2026-04-02
test: add post-deploy template assertion suite for integration fixtures


VirtueMe:
## Background

The current integration test suite (added in #745) deploys all fixtures to LocalStack and verifies that deployment succeeds. This catches runtime errors — malformed IAM policies, invalid state machine definitions, etc.

However, some bugs produce structurally incorrect CloudFormation templates that LocalStack accepts without error but that cause silent failures in production. Two examples already fixed:

- **#275** — multiple `AWS::SNS::TopicPolicy` resources targeting the same topic. CloudFormation silently overwrites the first with the second. LocalStack deploys green; the bug is only visible by inspecting the generated template.
- **#470** — `{ Ref: LambdaFunction }` in an IAM policy resource resolves to the function *name*, not its ARN. LocalStack's IAM validates this and rejects it — but only because IAM happens to validate it. A future similar bug might not be caught this way.

## Proposed solution

After `sls compose deploy` runs (which internally packages each fixture and writes the compiled template to `.serverless/cloudformation-template-update-stack.json`), run a second step that executes `verify.test.js` files co-located with each fixture:

```
fixtures/
  notifications/
    serverless.yml
    verify.test.js   ← asserts exactly one SNS TopicPolicy per topic, etc.
  circular-dependency/
    serverless.yml
    verify.test.js   ← asserts no { Ref: Lambda } in IAM policy resources, etc.
  basic-state-machine/
    serverless.yml
    # no verify.test.js needed — deploy success is sufficient
```

Runner added to the CI workflow after the deploy step:
```bash
npx mocha fixtures/**/verify.test.js
```

The `verify.test.js` files use the same Mocha + Chai setup as the rest of the test suite and read the already-generated `.serverless/cloudformation-template-update-stack.json` for their assertions.

## Why this matters

This closes the gap between "deploys without error" and "generates a correct template". It also gives future contributors a clear pattern for adding template-level assertions when fixing compilation bugs.

PR #802, opened 2026-07-31
fix(notifications): dedupe SNS/SQS policy statements by Sid (#801)

### LINKED ISSUE #801: Notifications: merged SNS/SQS policy has duplicate Sids → "Every policy statement must have a unique ID" (regression from #747)
## Summary

After #747 (merged), upgrading produces a CloudFormation/AWS deploy failure when two or more state machines send notifications to the **same** SNS topic or SQS queue:

```
Invalid parameter: Every policy statement must have a unique ID
```

Reported on the PR by @hfhchan-plb: https://github.com/serverless-operations/serverless-step-functions/pull/747#issuecomment-5140740807

## Root cause

#747 added `mergeResourcePolicies` in `lib/deploy/stepFunctions/compileNotifications.js`, which merges the per-state-machine `AWS::SNS::TopicPolicy` / `AWS::SQS::QueuePolicy` resources that target the same topic/queue into a single resource by **concatenating their statements**.

The problem: the statements it concatenates are byte-for-byte identical, including their `Sid`. The `Sid` is generated by `generatePolicyName` in `lib/deploy/stepFunctions/notificationStrategies/utils.js`:

```js
function generatePolicyName(status, targetType, action, resource) {
  const suffix = crypto.createHash('md5')
    .update(JSON.stringify({ action, resource }))   // only action + resource
    .digest('hex').substr(0, 5);
  return `${status}-${targetType}-${suffix}`;         // status + type + hash
}
```

The `Sid` depends only on **status + target-type + (action, resource)** — it carries **no state-machine identity**. So when machine A and machine B both notify topic X on `FAILED`, each emits a statement with an identical `Sid` (`FAILED-sns-<hash>`), identical `Principal` (`events.amazonaws.com`), `Effect`, `Action`, and `Resource`.

Merging stacks these identical statements into one policy, so the policy ends up with duplicate `Sid`s → AWS rejects the template with `Every policy statement must have a unique ID`.

## Why #747's tests didn't catch it

- The unit/verify tests only **counted** statements ("10 merged statements") and never asserted `Sid` uniqueness.
- LocalStack silently accepts duplicate `Sid`s where real CloudFormation/AWS does not — the same LocalStack blind spot tracked in #748.

## Proposed fix

Dedupe statements during the merge. Because same-target statements are fully identical, deduping by `Sid` (equivalently, full-statement equality) is both the correct fix for the crash **and** semantically correct: the policy grants EventBridge publish/send to the shared topic/queue, and only needs to state that once — it is not per-machine. Two machines sharing topic X → one statement per status, not two.

This preserves #747's goal (one merged policy per target, no silent overwrite) while producing a valid statement set.

## Acceptance criteria

- [x] Merged SNS/SQS policies contain no duplicate `Sid`s
- [x] Unit test asserting `Sid` uniqueness in a merged policy (fails before the fix)
- [x] `notifications` fixture `verify.test.js` asserts `Sid` uniqueness so the integration suite catches this class going forward

## Reproduction

Two state machines, both with a `notifications` block pointing at the same SNS topic (or SQS queue). Package and inspect the compiled `AWS::SNS::TopicPolicy` — the `PolicyDocument.Statement` array contains repeated `Sid`s.

Regression from #747. Related: #275, #748.

VirtueMe:
Closes #801

## Summary
- `mergeResourcePolicies` (added in #747) concatenates same-target SNS/SQS policy statements, but the `Sid` carries no state-machine identity — it's derived from status + target type + (action, resource) only. So two state machines sharing a topic/queue emit **byte-identical** statements with the same `Sid`. AWS rejects that template: `Invalid parameter: Every policy statement must have a unique ID` (regression from #747, reported on that PR).
- Dedupe statements by `Sid` while merging (new `mergeStatements` helper). The grant is per-target, not per-machine, so keeping one statement per `Sid` is correct. The SNS/SQS branches are folded into one table-driven loop.
- Unit tests assert `Sid` uniqueness for shared SNS/SQS targets; two pre-existing count assertions (10→5) that encoded the buggy pre-dedup behaviour are corrected.
- The `notifications` fixture `verify.test.js` swaps its statement-count checks for `Sid`-uniqueness guards. LocalStack accepts duplicate `Sid`s (which is how #747 shipped green), so only the compiled-template assertion catches this class — see #748.

## Test plan
- [x] `npm test` — 618 passing
- [x] `npm run lint` — clean
- [x] Packaged `notifications` fixture → one merged policy per target, 1 statement with a unique `Sid` each

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/deploy/stepFunctions/compileNotifications.js
```diff
@@ -199,40 +199,47 @@ function validateConfig(serverless, stateMachineName, stateMachineObj, notificat
...
+  }
+}
+
 function mergeResourcePolicies(resourcePairs) {
-  const snsMap = new Map(); // JSON-serialised Topics array → { logicalId, resource }
-  const sqsMap = new Map(); // JSON-serialised Queues array → { logicalId, resource }
+  // policyType → { keyOf(resource) → { logicalId, resource } }
+  const mergedByType = {
+    'AWS::SNS::TopicPolicy': { map: new Map(), keyOf: (r) => JSON.stringify(r.Properties.Topics) },
+    'AWS::SQS::QueuePolicy': { map: new Map(), keyOf: (r) => JSON.stringify(r.Properties.Queues) },
```
zirkelc: A stable key might require a sort first, otherwise the keys follow insertion order.

@copilot wdyt?

VirtueMe: Very good call!

lib/deploy/stepFunctions/compileNotifications.js
```diff
@@ -216,11 +216,15 @@ function mergeStatements(existing, incoming) {
   }
 }
 
+function policyTargetKey(targets) {
+  return JSON.stringify(_.sortBy([].concat(targets), (target) => JSON.stringify(target)));
```
zirkelc: @copilot that still does not sort the keys, right?

here is what I meant:

```ts
function stringifyJSON(obj) {
  return JSON.stringify(obj, (key, value) =>
    value && typeof value === 'object'
      ? Object.keys(value)
          .sort()
          .reduce<{ [key: string]: unknown }>((result, k) => {
            result[k] = value[k];
            return result;
          }, {})
      : value,
  );
}
```

zirkelc: @copilot+claude-opus-4.6 why is better in your solution than using the the replace arg of JSON.stringify?

zirkelc: @copilot yes do that

